PR #26, opened 2019-07-29
Upgrade to gradle 5.5.1

goober:
Replaces com.iadams.sonar-packaging plugin with com.github.johnrengelman.shadow since the former does not support gradle 5.x.
See https://github.com/iwarapter/gradle-sonar-packaging-plugin/issues/22


gradlew.bat
```diff
@@ -1,84 +1,84 @@
...
-:fail
-rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
-rem the _cmd.exe /c_ return code!
-if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
-exit /b 1
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
```
mc1arke: I'm guessing that this has been written on a *Nix Based OS and the line endings have changed. Given this file specifically targets windows, could you check that the line endings are CRLF please?

goober: I have confirmed that the gradlew.bat has CRLF line endings.
In addition I have now also added a `.gitattributes` file that will keep it that way when the project moves forward.

build.gradle
```diff
@@ -57,7 +59,7 @@ dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'org.mockito', name: 'mockito-core', version: '2.24.0'
     zip "sonarqube:sonarqube:${sonarqubeVersion}@zip"
-    compileOnly group: 'org.sonarsource.sonarqube', name: 'sonar-plugin-api', version: "7.7"
+    compileOnly group: 'org.sonarsource.sonarqube', name: 'sonar-plugin-api', version: "${sonarqubeVersion}"
```
mc1arke: I suspect this line can actually be removed. It was enforced by the `sonarPackaging` plugin stating that `sonar-plugin-api` is required, but the full SonarQube API/internals are already included through the `sonarLibraries` reference above.

goober: It worked fine without the dependency as well. PR updated

mc1arke:
Merged. Thanks for your contribution.

Issue #38, opened 2019-10-22
Upgrade to support sonarqube 8.0


thib3113:
**Describe the bug**
For some reason, we need to upgrade to 8.0 ( sonarqube db is destroy, and 8.0 repair it ... ). 
Did you plan to adapt the plugin for 8.0 ?

Here is the error from sonarqube :
```
java.lang.IllegalStateException: Fail to load plugin Community Branch Plugin [communityBranchPlugin]
	at org.sonar.server.plugins.ServerExtensionInstaller.installExtensions(ServerExtensionInstaller.java:88)
	at org.sonar.ce.container.ComputeEngineContainerImpl.startLevel4(ComputeEngineContainerImpl.java:230)
	at org.sonar.ce.container.ComputeEngineContainerImpl.start(ComputeEngineContainerImpl.java:196)
	at org.sonar.ce.ComputeEngineImpl.startup(ComputeEngineImpl.java:45)
	at org.sonar.ce.app.CeServer$CeMainThread.attemptStartup(CeServer.java:163)
	at org.sonar.ce.app.CeServer$CeMainThread.run(CeServer.java:141)
Caused by: java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "getClassLoader")
	at java.base/java.security.AccessControlContext.checkPermission(Unknown Source)
	at java.base/java.security.AccessController.checkPermission(Unknown Source)
	at java.base/java.lang.SecurityManager.checkPermission(Unknown Source)
	at java.base/java.lang.ClassLoader.checkClassLoaderPermission(Unknown Source)
	at java.base/java.lang.Class.getClassLoader(Unknown Source)
	at com.github.mc1arke.sonarqube.plugin.classloader.ClassReferenceElevatedClassLoaderFactory.createClassLoader(ClassReferenceElevatedClassLoaderFactory.java:47)
	at com.github.mc1arke.sonarqube.plugin.CommunityBranchPluginBootstrap.define(CommunityBranchPluginBootstrap.java:60)
	at org.sonar.server.plugins.ServerExtensionInstaller.installExtensions(ServerExtensionInstaller.java:78)
	... 5 common frames omitted
```

**To Reproduce**
1. Install sonarqube 8.0
2. Add this plugin in plugin folder
3. start sonarqube

**Expected behavior**
work

**Software Versions**
 - SonarQube Version: 8.0-beta
 - Plugin Version: 1.2.0-snapshot

**Additional context**
[...]

PR #41, opened 2019-10-24
#38: Add support for SonarQube 8.0

mc1arke:
Replaces references to the removed `ScannerUtils` class with Java's `URLEncoder`, and adds implementations of new methods introduced into SonarQube 8.0's interfaces. Since SonarQube 8.0 also changed `ScannerWsClient` from a class to an interface, the `ScannerWsClient` is now invoked through reflection in `ScannerWsClientWrapper` to prevent a `java.lang.IncompatibleClassChangeError` being thrown during invocation.

On top of this, SonarQube Compute Engine now contains a Java Security Manager that blocks access to retrieving `ClassLoader`s, so prevents the work-around used to access the branch API classes required for this plugin to execute. Moving the loading of the impacted classes into a `CoreExtension` and bypassing the class-loading manipulation for these classes overcomes this problem. This will require users of this plugin to install the plugin to both the plugins directory and the core platform's `lib/common` directory (or symlink the files) given plugin classes are only searched for from the plugins directory, and extensions are only loaded from the `common` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ SonarQube Version | Plugin Version
 7.4 - 7.7         | 1.0.2
 
 # Installation
-Either build the project or [download a compatible release version of the plugin JAR](https://github.com/mc1arke/sonarqube-community-branch-plugin/releases). Copy the plugin JAR file to the `extensions/plugins/` directory of your SonarQube instance and restart SonarQube.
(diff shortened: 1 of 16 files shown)


mateuszmroz93:
What about this PR?

PR #61, opened 2020-01-07
Distribution of the plugin as docker image

tisoft:
* build the plugin inside docker
* create a minimal image with the build jar
* added docker-compose file for quick startup

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,53 @@
+version: "2.1"
+
+services:
+  sonarqube:
+    depends_on:
+      - sonarqube-community-branch-plugin
+      - db
+    image: sonarqube:7.9-community
+    container_name: sonarqube
+    ports:
+      - 9000:9000
+    networks:
+      - sonarnet
+    environment:
+      - SONARQUBE_JDBC_URL=jdbc:postgresql://db:5432/sonar
+      - SONARQUBE_JDBC_USERNAME=sonar
+      - SONARQUBE_JDBC_PASSWORD=sonar
+    entrypoint:
+      - /sonarqube-community-branch-plugin/run.sh
+    volumes:
+      - sonarqube_conf:/opt/sonarqube/conf
+      - sonarqube_data:/opt/sonarqube/data
+      - sonarqube-community-branch-plugin:/sonarqube-community-branch-plugin
+
+  sonarqube-community-branch-plugin:
+    image: tisoft/sonarqube-community-branch-plugin
+    network_mode: none
+    build:
+      context: .
+    volumes:
+      - sonarqube-community-branch-plugin:/sonarqube-community-branch-plugin
+
+  db:
+    image: postgres:11
+    container_name: postgres
+    networks:
+      - sonarnet
+    environment:
+      - POSTGRES_USER=sonar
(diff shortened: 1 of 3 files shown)


tisoft:
I played around a bit on how to distribute the plugin as docker image, so that it can be easily combined with the stock sonarqube image. This is the best I could come up with.

It builds the plugin inside docker, and then creates a minimal image with only the resulting jar file and a wrapper script for the sonarqube entrypoint.

In the compose file you need to specify a shared volume and change the entrypoint of sonarqube to the file from the plugin image.

This should allow easy sonarqube and easy plugin updates. You can update both at the same time or one after another.

The only other idea to have an easy to manage configuration would be to extend the sonarqube image and copy the plugin inside, but that would need an update to the sonarqube-with-this-plugin image everytime a new sonar image is released.

nixel2007:
https://github.com/Daabramov/Sonarqube-for-1c-docker/blob/master/8.0-ce-beta.Dockerfile

Dockerfile based on official docker image with curl'ed installation.

mc1arke:
> that would need an update to the sonarqube-with-this-plugin image everytime a new sonar image is released.

Given every SonarQube release so far has required changes to the plugin to be compatible with SonarQube API changes, this restriction doesn't seem like a major issue, and shouldn't need users to change the entrypoint.

4n4n4s:
I agree with @nixel2007 here the easiest thing possible would be to use an existing image. 
Let's say we use "sonarqube:lts" as our base image what I would do in the dockerfile only is: 
- curl for the specified version of the plugin using the [releases](https://github.com/mc1arke/sonarqube-community-branch-plugin/releases) 
- copy the plugin into the required locations

And on dockerhub enable autobuild so that whenever the base image is changing we're also getting the updated image built.

tisoft:
Better this way?

I used the explicit version 7.9 instead of lts in the hope that, that will stay compatible. lts can be switched to a newer version and render the plugin unusable.

Also I build the plugin instead of fetching it with curl or wget, so that the dockerfile is not dependent on any release build. In fact you can use it to build your local versions, with all modifications you have in your local working directory.

To be usefull this repo would need to be added to dockerhub autobuilds, so that a new push to master (or a new tag) creates a new image version.

What do you think?

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "2.1"
+
+services:
+  sonarqube:
+    depends_on:
+      - db
+    image: sonarqube-with-community-branch-plugin:7.9-community
```
tisoft: This would need to be renamed to the official docker image name, if the repo is added to dockerhub

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+    build:
+      context: .
```
tisoft: I've added the build key, so that you can do

    docker-compose build
    docker-compose up

to easily test any local changes of the plugin.

marvin-w:
The approach from @4n4n4s is what you'd normally do and what is the preferred way in such cases - at least in my opinion.

mc1arke:
I'm proposing to supercede this with #312. Any thoughts?

tisoft:
Closed in favor of #312.

Issue #66, opened 2020-01-13
need a global default value for sonar.pullrequest.provider


jizhilong:
Most organization/company  host their  repositories with a single ALM(actually I just know ALM is github like tools, but not its full name, could anyone tell me?),  a global default value for  `sonar.pullrequest.provider` would reduce a lot of repeat configuration work.
I can contribute a PR if this feature is compatible with the original design.

tisoft:
ALM its application lifecycle management. https://en.m.wikipedia.org/wiki/Application_lifecycle_management

mc1arke:
There's no specific pattern we're required to follow for PR decoration, so I'm happy with what @tisoft has proposed in his Pull Request. However, each project would still need to have it's repository name/path/slug set individually, do you have any ideas about how to prevent that needing done manually per-project?

tisoft:
We could think about making the name/path/slug part customizable with parameters:

`${user}/${project}`

would need to find some sensible parameters we could use, eg.

* user: the username associated with the API token
* project: the project name of the sonar project

The other possibility would be, if the name/path/slug option could also be set on the scanner side. Even better, if all options could be set on the scanner side, you would need no config at all in sonar itself, it could all be done from the build system.

I can't think of any other general way, to solve the issue for name/path/slug.

jizhilong:
+1 for making repositorySlug configurable on the scanner side.

PR #68, opened 2020-01-13
Make common options configurable globally

tisoft:
fixes: #66


tisoft:
In my opinion this is how we could split the configuration options, that can be configured globally and locally from those that only make sense locally to a project. What do you think?

PR #69, opened 2020-01-13
loose qualifier for property sonar.pullrequest.provider

jizhilong:
implement #66


jizhilong:
> Thanks for the contribution. I may merge #68 in preference to this given it covers a slightly wider set of cases

OK, I'll close this PR then.

PR #74, opened 2020-01-14
Fix adding discussions for gitlab

uwej711:
Hi, 

thanks for this great plugin. I did some tests decorating merge request with our gitlab server and had to change this to make it work.

1. always get the commits regardless of deleteCommentsEnabled (otherwise no discussion is created below)
2. add old_path when adding the discussion (otherwise the comment is not displayed inline in the diff in gitlab)

Kind regards,
Uwe

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabServerPullRequestDecorator.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabServerPullRequestDecorator.java
@@ -126,7 +126,7 @@ public class GitlabServerPullRequestDecorator implements PullRequestBuildStatusD
             User user = getSingle(userURL, headers, User.class);
             LOGGER.info(String.format("Using user: %s ", user.getUsername()));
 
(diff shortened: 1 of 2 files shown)


tisoft:
Hi,

>     1. always get the commits regardless of deleteCommentsEnabled (otherwise no discussion is created below)
Thanks, that is clearly a bug!

>     2. add old_path when adding the discussion (otherwise the comment is not displayed inline in the diff in gitlab)
In my tests this was not necessary. I will try to debug when the old_path is really necessary.

uwej711:
I tried this with changes to an existing file and by adding a new file and I needed old_path in both cases, on a self hosted Gitlab FOSS 12.6.2.

Without it the comments where not directly visible in the diff. Sometimes they showed up when running another analysis with new commits on the same merge request but they never showed up there on the first analysis. This might also be a gitlab issue.

mc1arke:
@tisoft Did you get a chance to investigate the second point reported in this change?

tisoft:
Not yet. Will do it this week.

kortov:
related to #86

mc1arke:

PR #81, opened 2020-01-26
#80 - Serve PR decoration images as static resources

### LINKED ISSUE #80: Serve PR decoration images as static resources
**Describe the bug**
In order to display the PR decoration icons in the notifications created by the plugin, it is currently necessary to either have Internet access to https://raw.githubusercontent.com/mc1arke/sonarqube-community-branch-plugin/master/src/main/resources/pr-decoration-images or to host the images on a private web server and configure the `com.github.mc1arke.sonarqube.plugin.branch.image-url-base` property.

**Expected behavior**
All of this should not be necessary. SonarQube allows plugin to [serve static resources ](https://docs.sonarqube.org/8.0/extend/developing-plugin/#header-2) as follows:

_If you need to serve static resources from your plugin such as images or JavaScript files, place them in a directory under resources named static (myplugin/src/main/resources/static). At runtime they'll be available from http://{server}/static/{pluginKey}/{file}._

**Software Versions**
 - SonarQube Version: 8.0
 - Plugin Version: 1.2.1-SNAPSHOT (current master)

don-vip:
Fixes #80 

Removes the `com.github.mc1arke.sonarqube.plugin.branch.image-url-base` property and use SonarQube static resources system instead.


mc1arke:
Thanks for your contribution. I had provided the property to allow people to be able to use my publically hosted images without having a need to keep the Sonarqube server online or visible 24/7, but I suspect this is an unlikely use-case so have no problem with removing it.

I've not run your updates yet, but the change looks good from a read-through. One observation I do have is around your commit message. I generally follow the rules in https://github.com/RomuloOliveira/commit-messages-guide/blob/master/README.md for writing a commit message. If you're feeling you need to put more in the Pull Request summary, then this probably indicates that something is missing from your commit message. If your message also doesn't make it why you're doing the change (not just what you've changed) then this normally indicates some missing content. Any chance you could update your commit message?

don-vip:
> Any chance you could update your commit message?

Sure! Done :)

mc1arke:

PR #58, opened 2019-12-20
SonarQube 8.1 Support

mc1arke:
Adds support for Sonarqube 8.1. As this change is not completely backwards compatible, the approach taken here may be introducing more complexity than required, so may be subject to change. In the meantime, this PR should be backwards compatible with Sonarqube 7.8 to 8.0 (inclusive)

**Please see the latest comments on this PR: the recent changes mean that this PR will no longer support SonarQube 8.0 and any older versions**

PR #122, opened 2020-03-12
Removed exception "IllegalStateException" if could not find New Coverage Condition in analysis and replaced to "0"

Grintek:
When in Sonar conditions there is no new_coverage condition (and I doubt that this condition is necessary) the plugin throws Exception. IllegalStateException("Could not find New Coverage Condition in analysis")


Grintek:
When in Sonar conditions there is no new coverage condition (and I doubt that this condition is necessary) the plugin throws Exception. IllegalStateException("Could not find New Coverage Condition in analysis")

exxbrain:
Looks like a kind of fix of #58

mc1arke:

Issue #106, opened 2020-02-26
Summary comment must be resolved by a hand before pull request could be merged.


exxbrain:
Summary comment must be resolved by a hand before pull request could be merged. So now pull request decoration at least in gitlab negatively affects a merge process when pull request couldn't be merged automatically after successful completion of a pipeline.

1. Navigate to your gitlab project’s settings page, select the **Only allow merge requests to be merged if all threads are resolved** check box and hit Save for the changes to take effect.
2. Make merge request without issues and click a button "Merge when pipeline succeeds"

**Expected behavior**
Pull request should be automatically merged.

**Screenshots**
<img width="957" alt="Screenshot 2020-02-26 at 14 06 47" src="https://user-images.githubusercontent.com/1011558/75339584-d84b1680-58a1-11ea-841b-e6e67a893918.png">

<img width="856" alt="Screenshot 2020-02-26 at 14 13 34" src="https://user-images.githubusercontent.com/1011558/75339780-3546cc80-58a2-11ea-8e6f-b34910fee8f4.png">

<img width="994" alt="Screenshot 2020-02-26 at 13 55 20" src="https://user-images.githubusercontent.com/1011558/75340631-e306ab00-58a3-11ea-96ff-8ab3a2c6a560.png">

**Software Versions**
 - Gitlab
 - SonarQube Version: 7.9.2 (build 30863)
 - Plugin Version: 1.2.1-RC3

**Suggestion**
I think that it will be good to resolve summary comment thread automatically if there are no issues found.

mc1arke:
This sounds like a feature request rather than a bug: you've enabled decoration, but your current pipeline then prevents decorated MRs from being auto-merged. I presume you're asking for the summary comment to posted in another way?

kortov:
I guess it's similar to #97

exxbrain:
> I presume you're asking for the summary comment to posted in another way?

If it is possible then yes. But current solution is nice as well except the case when there are no issues.

exxbrain:
> I guess it's similar to #97

@kortov I think you are right. But I posted a bug because current functionality is conflicting with gitlab settings.

exxbrain:
Gitlab provides 2 endpoints: discussions and notes. Now we use discussions one. https://docs.gitlab.com/ee/api/discussions.html, https://docs.gitlab.com/ee/api/notes.html. I think we can try notes.

kortov:
[...]

PR #124, opened 2020-03-13
if duplication and issues equal zero, to do repost "notes" instead "discussions" in Gitlab

Grintek:
If duplication and issues equal zero, to do repost "notes" instead "discussions" in GitLab. Also changed if in Sonar conditions there is no new coverage condition


exxbrain:
This pull request fixes #106.

PR #131, opened 2020-03-17
Modify to limit the number of annotations

choi-inkyun:
If there are more than 50 annotations, even if the response is 200, it is not exposed to Github Checks.
After limiting to 50, I confirmed that it works normally.

## reference
https://developer.github.com/v3/checks/runs/#parameters
image

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java
@@ -65,6 +65,8 @@ public class GraphqlCheckRunProvider implements CheckRunProvider {
     public static final String PULL_REQUEST_GITHUB_APP_ID = "sonar.alm.github.app.id";
     public static final String PULL_REQUEST_GITHUB_APP_NAME = "sonar.alm.github.app.name";
 
+    private static final int GITHUB_CHECKS_ANNOTATIONS_MAX_NUMBER_PER_REQUEST = 50;
+
     private static final Logger LOGGER = Loggers.get(GraphqlCheckRunProvider.class);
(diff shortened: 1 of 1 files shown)


kortov:
Why limit not list partition?)

choi-inkyun:
Please help me because I am new to java : )

kortov:
@choi-inkyun could you provide info what happens, when there are more than 50 annotations? Github returns 200 with no errors, but do not post them?

choi-inkyun:
@kortov 
I had an issue where the Decorator App was not visible on Github Checks. When the annotation exceeded 50, When I checked the response through ce.log, the response came as 200 and there was no error. However, nothing was displayed in the Github Checks. I've tried a number of things and found a case where the Decorator App is not visible due to this problem.

kortov:
Maybe you're right with just limiting annotations. I've tried enhance your PR in #132 but it seems it's just complicates the code. But you can test it if you'd like,

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java
```diff
@@ -104,6 +106,7 @@ public void createCheckRun(AnalysisDetails analysisDetails, UnifyConfiguration u
         headers.put("Accept", "application/vnd.github.antiope-preview+json");
 
         List<InputObject<Object>> annotations = analysisDetails.getPostAnalysisIssueVisitor().getIssues().stream()
+                .limit(GITHUB_CHECKS_ANNOTATIONS_MAX_NUMBER_PER_REQUEST)
```
mc1arke, line 109: Should we not be making multiple calls to the endpoint to submit all the annotations, rather than just submitting a subset? At the very least, I'd expect us to warn the user we've only submitted a subset of the findings

PR #143, opened 2020-03-25
Allow override `sonar.alm.github.app.name` from CLI

Brialius:
Implement #144

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java
@@ -95,7 +95,7 @@ public class GraphqlCheckRunProvider implements CheckRunProvider {
         String apiPrivateKey = unifyConfiguration.getRequiredServerProperty(PULL_REQUEST_GITHUB_TOKEN);
         String projectPath = unifyConfiguration.getRequiredProperty(PULL_REQUEST_GITHUB_REPOSITORY);
         String appId = unifyConfiguration.getRequiredServerProperty(PULL_REQUEST_GITHUB_APP_ID);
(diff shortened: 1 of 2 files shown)


Brialius:
@mc1arke This is really simple but useful enhancement, take a look please.

mc1arke:
@Brialius this change is no longer compatible with the version of the plugin from master given `UnifyConfigration` no longer exists. If you'd still like this functionality then please update your commit and ensure you detail why this change it required.

I'm personally not convinced that allowing override of the app name from the scanner is a good thing as it means 2 identical scan results can lead to different decoration, and potentially allows a malicious user to deface the target Github instance.

Brialius:
Sure @mc1arke, I'll take a look, thanks for a heads up.

Brialius:
Just to clarify my use case:
I have application code and regression tests code in the single repo and want to run 2 separate Sonarqube checks without overwriting results of each other in GH Check section (illustration is here: #144). 

To do so in 1.3.0 I should be able to change GH Check name via CLI or have two or more separate Sonqrqube instances. 
Property responsible for GH Check name was called as `appName` https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/143/commits/08dab1cdb6ab7b8ba2a74bf041fa5505d33b1edf# and used only for GH check name. 
Honestly I don't understand how it could be used to deface the target Github instance or other harmful action.

As I can see here https://github.com/mc1arke/sonarqube-community-branch-plugin/blob/master/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java#L130 "name" property is not configurable anymore. 
So now there is no way to have more then one separate GH Check per repo.

@mc1arke Do you see any value in this feature or it looks like very specific use case?

mc1arke:
@Brialius I'm not convinced the value this feature provides to other users given the specialized use case you seem to have. Given the lack of other people's reactions on your issue or this PR, I'm going to close them both, but you're obviously welcome to keep using them from your fork, and are free to open a new PR in the future if you feel more people are interested in this usage pattern.

PR #145, opened 2020-03-26
Correctly report coverage back to Gitlab

teake:
GitLab does not have a concept of 'new' coverage, only that of total coverage.
As such, SonarQube should report back the total coverage.

In addition, reporting coverage back is optional, see
https://docs.gitlab.com/ee/api/commits.html#post-the-build-status-to-a-commit.
So if there is no coverage information, we do not report it back.


teake:
Note that I haven't updated the tests accordingly, so they might fail. I have tested it on my SonarQube instance, and it seems to work correctly.

mc1arke:
Please rebase this onto the latest head of your target branch

teake:
Not a big fan of the whoe rebase + force push -> rewriting history thing, so I've created a new PR to supersede this one: #157.

Issue #134, opened 2020-03-18
PR decoration: For line type REMOVED the only valid file type is FROM (was: TO)


racodond:
Hi,

One of my Bitbucket PR decorations failed with the following error message:

`2020.03.18 21:16:50 INFO  ce[AXDvSpwAOt-M153AClS0][c.g.m.s.p.c.p.b.s.BitbucketServerPullRequestDecorator] starting to analyze with com.github.mc1arke.sonarqube.plugin.ce.pullrequest.AnalysisDetails@41982f12
2020.03.18 21:16:50 ERROR ce[AXDvSpwAOt-M153AClS0][c.g.m.s.p.c.p.b.s.BitbucketServerPullRequestDecorator] {"errors":[{"context":null,"message":"For line type REMOVED the only valid file type is FROM (was: TO)","exceptionName":"com.atlassian.bitbucket.validation.ArgumentValidationException"}]}`

Here's the diff of the PR:
![image](https://user-images.githubusercontent.com/1215713/77004749-ec91a900-695f-11ea-9a34-2ad96e674664.png)

I'm running SonarQube 7.9.2 and version 1.3.0 of the plugin.

Thank you

Regards,

David

4n4n4s:
Hi @racodond are you sure that the change in this file causes the issue? Cause it says line type "REMOVED" and this looks only like a change. 

Can you isolate it and only start a review with this change?

If the issue is still occuring I'll check the logic and push a fix for it. Thank you.

racodond:
Hi @4n4n4s,

I confirm the issue. I just create a project with one single file (see html code below).

![image](https://user-images.githubusercontent.com/1215713/77553097-a767fc80-6eb4-11ea-8ac6-5988392780ec.png)

It doesn't seem to be just a change. It's line removal + line addition (both of them marked as line #12).

`
<div class="todo-content">
    <h1 class="page-title">My Todos</h1>
    <div class="todo-create">
        <form #todoForm="ngForm" (ngSubmit) = "createTodo(todoForm)" novalidate>
            <input type="text" id="title" class="form-control" placeholder="Type a todo and press enter..."
                   required
                   name="title" [(ngModel)]="newTodo.title"
                   #title="ngModel" >
            <div *ngIf="title.errors && title.dirty"
                 class="alert alert-danger">
                <div [hidden]="!title.errors.required">
                    Title is required.
                </div>
            </div>
        </form>
    </div>
</div>
`

Thanks!

[...]

PR #148, opened 2020-03-31
Update documentation

4n4n4s:
Attempt to resolve. #129 #146 

Documentation - Ain't nobody got time for dat.

PR #149, opened 2020-03-31
Bitbucket fix removed line error

4n4n4s:
Fixing the error that occured when a line is reported as REMOVED and ADDED like in case of #134

Solution is to filter "REMOVED" entries completely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,33 +16,21 @@ SonarQube Version | Plugin Version
 7.8 - 8.0         | 1.3.0
 7.4 - 7.7         | 1.0.2
 
-# Installation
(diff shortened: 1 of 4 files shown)


mc1arke:
This seems to include your commit from #148 . Any chance you could drop it from this PR so I can review it independently?

mc1arke:
Any chance of resolving the conflict?

mc1arke:
Closing this since the whole Bitbucket implementation has been been changed to use Code Insights. I've not checked if the new implementation still contains this issue, but this fix will no loner be compatible with it.

PR #150, opened 2020-04-01
Add images base URL property

### LINKED ISSUE #130: [GitHub][PR Decoration] Private Sonarqube server static images unavailable
**Describe the bug**
Github uses `https://camo.githubusercontent.com/` to proxy any user images on their side and this [cannot be disabled](https://stackoverflow.com/questions/57857193/github-image-without-camo). It means every image should be available for GitHub proxy server, which is impossible with the way of serving static images #80 for private hosted Sonarqube server behind a strict firewall.

**To Reproduce**
1. Setup CE Branch plugin on a private Sonarqube server
2. Use GitHub as a provider
3. Run Sonarqube branch check

**Expected behavior**
Static images are available in GitHub PR checks for private and public accessible Sonarqube servers

**Screenshots**
image

image

**Software Versions**
 - SonarQube Version: Version 7.9.2 (build 30863)
 - Plugin Version: 1.3.0

**Additional context**
Please return back function of `com.github.mc1arke.sonarqube.plugin.branch.image-url-base` and make it optional or provide other way to solve this issue.

Brialius:
Fix #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,13 @@ To enable setting of several properties in SonarQube on project level is require
 
 The property "projectKey" or "userSlug" are mandatory in order to decide which API endpoint should be used.
 
+## Serving images for PR decoration
+By default, images for PR decoration are served as static resources on the SonarQube server as a part of Community Branch Plugin. 
+
+If you use a SonarQube server behind a firewall and/or PR service (Github, Gitlab etc..) hasn't access to SonarQube server, you should change `Images base URL` property in `General > Pull Request` settings.
+
+Anyone needing to set this value can use the URL `https://raw.githubusercontent.com/mc1arke/sonarqube-community-branch-plugin/master/src/main/resources/static`, or download the files from this location and host them themself.
+
 # Contribution
 To generate the jar file to copy to your Sonar Server execute ```./gradlew clean build``` inside of the project dir. This will put the jar under ```libs/sonarqube-community-branch-plugin*.jar```
 
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/PullRequestPostAnalysisTask.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/PullRequestPostAnalysisTask.java
@@ -125,7 +125,7 @@ public class PullRequestPostAnalysisTask implements PostProjectAnalysisTask,
                                     postAnalysisIssueVisitor, qualityGate,
                                     new AnalysisDetails.MeasuresHolder(metricRepository, measureRepository,
                                                                        treeRootHolder), analysis,
-                                    projectAnalysis.getProject(), server.getPublicRootUrl());
+                                    projectAnalysis.getProject(), configuration, server.getPublicRootUrl());
 
         PullRequestBuildStatusDecorator pullRequestDecorator = optionalPullRequestDecorator.get();
         LOGGER.info("using pull request decorator" + pullRequestDecorator.name());
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
@@ -46,6 +46,7 @@ import org.sonar.core.extension.CoreExtension;
 public class CommunityBranchPlugin implements Plugin, CoreExtension {
 
     public static final String PULL_REQUEST_PROVIDER = "sonar.pullrequest.provider";
+    public static final String IMAGE_URL_BASE = "com.github.mc1arke.sonarqube.plugin.branch.image-url-base";
     private static final String PULL_REQUEST_CATEGORY_LABEL = "Pull Request";
     private static final String GITHUB_INTEGRATION_SUBCATEGORY_LABEL = "Integration With Github";
     private static final String GENERAL = "General";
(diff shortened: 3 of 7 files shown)


Brialius:
@don-vip please take a look

Brialius:
Example of base URL: https://raw.githubusercontent.com/mc1arke/sonarqube-community-branch-plugin/master/src/main/resources/static

Brialius:
> Please also document the property in README. I removed it, it must be restored.

Thanks for review!

Instead of previous approach I made new property configurable via server UI with appropriate field description: https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/150/files#diff-6b77d55ef12cf09f58f256a700f17effR93-R97

But you are right, maybe an explanation in the README could be useful to get the clue about new property usage.

4n4n4s:
Hi @Brialius @don-vip 
so adding https://raw.githubusercontent.com/SonarSource/sonarcloud-github-static-resources/gh-pages/v2/ as base URL should work fine right?

Brialius:
> Hi @Brialius @don-vip
> so adding https://raw.githubusercontent.com/SonarSource/sonarcloud-github-static-resources/gh-pages/v2/ as base URL should work fine right?

Hey @4n4n4s 
I suppose any storage with the [same folders structure](https://github.com/mc1arke/sonarqube-community-branch-plugin/tree/master/src/main/resources/static) and images inside should work fine.

Pls notice that there is no trailing slash in base URL

4n4n4s:
> Pls notice that there is no trailing slash in base URL

Maybe we should remove it in case someone configures it with a slash to prevent misconfiguration

Brialius:
> > Pls notice that there is no trailing slash in base URL
> 
> Maybe we should remove it in case someone configures it with a slash to prevent misconfiguration

@4n4n4s 
fixed

Brialius:
@don-vip property description to README is added

Brialius:
@mc1arke could you please take a look at this PR?

README.md
```diff
@@ -30,6 +30,13 @@ To enable setting of several properties in SonarQube on project level is require
 
 The property "projectKey" or "userSlug" are mandatory in order to decide which API endpoint should be used.
 
+## Serving images for PR decoration
+By default, images for PR decoration are served as static resources on the SonarQube server as a part of Community Branch Plugin. 
+
+If you use a SonarQube server behind a firewall and/or PR service (Github, Gitlab etc..) hasn't access to SonarQube server, you should change `Images base URL` property in `General > Pull Request` settings.
+
+It is used to load the images for the PR decoration. [Default public base images URL](https://raw.githubusercontent.com/mc1arke/sonarqube-community-branch-plugin/master/src/main/resources/static).
```
mc1arke, line 38: This line seems confusing. I suspect you're trying to say that `Anyone needing to set this value can use the URL 'https://raw.githubusercontent.com/mc1arke/sonarqube-community-branch-plugin/master/src/main/resources/static', or download the files from this location and host them themself`.

I also have a slight concern about pointing people to master, as I've then either got to keep any old images on master (and therefore included in each plugin release), or risk breaking rendering of user's pull requests if the images in this plugin ever change. I'd potentially ask users to load images off the tag related to their release, but don't know if that over-complicates the setup.

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/PullRequestPostAnalysisTask.java
```diff
@@ -125,7 +125,7 @@ public void finished(PostProjectAnalysisTask.ProjectAnalysis projectAnalysis) {
                                     postAnalysisIssueVisitor, qualityGate,
                                     new AnalysisDetails.MeasuresHolder(metricRepository, measureRepository,
                                                                        treeRootHolder), analysis,
-                                    projectAnalysis.getProject(), server.getPublicRootUrl());
+                                    projectAnalysis.getProject(), configuration, server.getPublicRootUrl());
```
mc1arke, line 128: This class is getting too big judging by the number of constructor parameters. There probably needs to be some refactoring of this class, but I'm happy to pick this up at a later point if needed.

README.md
Brialius: @mc1arke Thanks for review, I've updated README according your recommendation.

Regarding your doubts about pointing to master, unfortunately I see no easy and universal solution to handle possible breaking changing for images hosting without increasing complexity and/or decreasing of flexibility.

\# | Solution | Possible issues | Comment
---|-----------|-----------------|------------
1 | Use static resources | **No issues** | Images are always actual
2 | Input field for URL |  | Manual input
2.1 | Github URL pointing to master | Risk of breaking rendering | Maintainer should keep all previous version of images in master to provide backward compatibility
2.2 | Github URL pointing to tag | Risk of breaking rendering | User should keep this link in-sync with installed version of plugin
2.3 | Self-hosted images  | Risk of breaking rendering | User should keep them in-sync with installed version of plugin
3 | Additional config switch, like `Use official public images URL: yes/no` | **No issues** | Plugin will generate a proper images URL using actual release tag by itself (not clear how to make it works for unreleased snapshots and PRs)

1+2 most simple and pretty flexible
1+3 most reliable but not flexible
1+2+3 flexible but complex from my prospective

Brialius: @mc1arke Could you please provide any suggestion about this?

mc1arke: Sorry for the delayed response. I'm happy to stick with your changes as-is for now.

PR #153, opened 2020-04-12
fix(binding_action): sonarqube api break change for select by project

ndeitch:
Hi,

I'm running sonarqube **Version 8.2 (build 32929)** and was getting 500 error when trying to set up decorator on project.

The error was `No such method error` then I checked and in this [commit](https://github.com/SonarSource/sonarqube/commit/ace9a50d55d831ec71b7db421aa04d1198392c6c#diff-86b128d8b719827a62a3a81bfb771993L64-R66) on sonarqube the method signature was changed. This fix change method invocation.

This pr should be merged in this branch? Because, I think will break 8.1 support, right?


mc1arke:
I believe this is superseded by #156. If it isn't please rebase your MR onto the updated target branch so it can be reviewed.

ndeitch:
You're right, thanks!

Issue #112, opened 2020-02-27
1.3.0 plugin works with sonar 8.2 ?


ktibi:
Hello,

We try to deploy the plugin on sonarqube 8.2 but we have a trace : 

```
......
INFO: Load project branches
INFO: Load project branches (done) | time=23ms
INFO: Load project pull requests
INFO: Load project pull requests (done) | time=20ms
INFO: Load branch configuration
INFO: ------------------------------------------------------------------------
INFO: EXECUTION FAILURE
INFO: ------------------------------------------------------------------------
INFO: Total time: 2.268s
INFO: Final Memory: 9M/40M
INFO: ------------------------------------------------------------------------
ERROR: Error during SonarQube Scanner execution
java.lang.IllegalStateException: Unable to load component class org.sonar.scanner.scan.filesystem.InputComponentStore
	at org.sonar.core.platform.ComponentContainer$ExtendedDefaultPicoContainer.getComponent(ComponentContainer.java:65)
	at org.picocontainer.DefaultPicoContainer.getComponent(DefaultPicoContainer.java:632)
	at org.picocontainer.parameters.BasicComponentParameter$1.resolveInstance(BasicComponentParameter.java:118)
	at org.picocontainer.parameters.ComponentParameter$1.resolveInstance(ComponentParameter.java:136)
	at org.picocontainer.injectors.SingleMemberInjector.getParameter(SingleMemberInjector.java:78)
	at org.picocontainer.injectors.ConstructorInjector$CtorAndAdapters.getParameterArguments(ConstructorInjector.java:309)
	at org.picocontainer.injectors.ConstructorInjector$1.run(ConstructorInjector.java:335)
	at org.picocontainer.injectors.AbstractInjector$ThreadLocalCyclicDependencyGuard.observe(AbstractInjector.java:270)
	at org.picocontainer.injectors.ConstructorInjector.getComponentInstance(ConstructorInjector.java:364)
	at org.picocontainer.injectors.AbstractInjectionFactory$LifecycleAdapter.getComponentInstance(AbstractInjectionFactory.java:56)
	at org.picocontainer.behaviors.AbstractBehavior.getComponentInstance(AbstractBehavior.java:64)
	at org.picocontainer.behaviors.Stored.getComponentInstance(Stored.java:91)
	at org.picocontainer.DefaultPicoContainer.instantiateComponentAsIsStartable(DefaultPicoContainer.java:1034)
[...]

PR #155, opened 2020-04-18
[GITLAB] Support proxy by using HttpClients.createSystem()

r-oldenburg:
Tested this on local (docker/kubernetes) setup. With this diff gitlab MR decoration began to work as it now honors the system properties for proxy. (http.proxyHost, http.proxyport, etc.)

Came up initially when setting it up with sonar 8.2. See [here](https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/112#issuecomment-614627627).


r-oldenburg:
@mc1arke: something wrong?

mc1arke:
Merged, thanks for the contribution.

PR #157, opened 2020-04-20
Correctly report coverage back to Gitlab

teake:
GitLab does not have a concept of 'new' coverage, only that of total coverage.
As such, SonarQube should report back the total coverage.

In addition, reporting coverage back is optional, see
https://docs.gitlab.com/ee/api/commits.html#post-the-build-status-to-a-commit.
So if there is no coverage information, we do not report it back. (Otherwise GitLab will erroneously report 0% coverage).

Note that I haven't updated the tests accordingly, so they might fail. I have tested it on my SonarQube instance, and it seems to work correctly.


src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabServerPullRequestDecorator.java
```diff
@@ -80,6 +80,9 @@
     public static final String PULLREQUEST_GITLAB_PIPELINE_ID =
             "com.github.mc1arke.sonarqube.plugin.branch.pullrequest.gitlab.pipelineId";
 
+    private static final String NO_COVERAGE_VALUE =
+            "-1";
```
mc1arke, line 84: The new-line doesn't seem right - this should be on the line above

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabServerPullRequestDecorator.java
```diff
@@ -155,10 +158,10 @@ public void decorateQualityGateStatus(AnalysisDetails analysis, AlmSettingDto al
                 }
             }
 
-            String coverageValue = analysis.findQualityGateCondition(CoreMetrics.NEW_COVERAGE_KEY)
+            String coverageValue = analysis.findQualityGateCondition(CoreMetrics.COVERAGE_KEY)
                     .filter(condition -> condition.getStatus() != QualityGate.EvaluationStatus.NO_VALUE)
                     .map(QualityGate.Condition::getValue)
-                    .orElse("0");
+                    .orElse(NO_COVERAGE_VALUE);
```
mc1arke, line 164: Why are we setting it to a value, rather than `null` or potentially just remaining as an optional with a subsequent `hasValue` check?

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabServerPullRequestDecorator.java
teake: The coding converting wasn't clear to me from the previous lines. Feel free to change it though.

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabServerPullRequestDecorator.java
teake: Good question. My Java is a little bit rusty so I went for this option, but I'm happy to take suggestions here.

mc1arke: I'll check where the variable is referenced and get you a suggestion shortly

mc1arke:
The intent of the first part of this MR doesn't seem right. Gitlab doesn't have the concept of new coverage, but we want to report the coverage of lines affected by this MR rather than the whole project, which I believe the existing code is doing. Am I incorrect in my understanding here?

teake:
> The intent of the first part of this MR doesn't seem right. Gitlab doesn't have the concept of new coverage, but we want to report the coverage of lines affected by this MR rather than the whole project, which I believe the existing code is doing. Am I incorrect in my understanding here?

Well, yes and no. GitLab only has the concept of test coverage for the whole project. So when the CI runner parses coverage from a job (see https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-parsing), that means coverage for the whole codebase on that branch.

While GitLab can show coverage on a MR, this also has the meaning of "coverage for the whole codebase in the source branch". There's an open issue for adding /displaying coverage for just the diff (https://gitlab.com/gitlab-org/gitlab/-/issues/20895), but this is not yet implemented.

So while it might be possible for SonarQube to report back the coverage of just the diff, we should not do this. In GitLab coverage always means coverage of the whole codebase, and reporting something else is confusing at least and misleading at worst. Until GitLab has a native capability to distinguish and display coverage of diffs, we should just report back coverage of the whole codebase IMHO.

teake:
I've made a few additional changes. Since coverage is not always a condition on a Quality Gate, it's better to get its information from a Measure. Also, coverage is not always present, so it should be optionally included in the summary report.

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabServerPullRequestDecorator.java
teake: Still waiting on feedback for this one.

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/AnalysisDetails.java
```diff
@@ -116,20 +116,20 @@ public String getCommitSha() {
...
-        BigDecimal newCoverage =
-                findQualityGateCondition(CoreMetrics.NEW_COVERAGE_KEY)
-                    .filter(condition -> condition.getStatus() != EvaluationStatus.NO_VALUE)
-                    .map(QualityGate.Condition::getValue)
-                    .map(BigDecimal::new)
-                    .orElse(null);
+        BigDecimal newCoverage = findMeasure(CoreMetrics.NEW_COVERAGE_KEY)
+                .map(Measure::getDoubleValue)
+                .map(BigDecimal::new)
+                .orElse(null);
```
teake, line 122: Ok, this doesn't work; it throws `java.lang.IllegalStateException: null`. But the previous way to get to get the new coverage value (via `findQualityGateCondition`) always returns 0% even in cases where it isn't.

@mc1arke Do you have any idea what the correct way to get the new coverage percentage?

teake: To clarify: getting the new coverage via `findQualityGateCondition` will yield 0% if coverage is not a condition on new code. If it is, the it will return the correct value.

PR #151, opened 2020-04-06
Add support for Bitbucket Code Insights

goober:
This change will add support for Bitbucket Code Insights in favor of regular comments when available. It will fall back on the comments strategy when the Code Insights is not available (it is supported in version 5.15 and later).

PR #158, opened 2020-04-20
SonarQube 8.2 support with BitBucket Code Insights

vklindukh:
This PR is based on https://github.com/andrejpetras/sonarqube-community-branch-plugin/tree/sonarqube82 branch with addition change, which implements BitBucket Code Insights functionality https://developer.atlassian.com/server/bitbucket/how-tos/code-insights/


goober:
Great contribution @vklindukh. However there is already an open PR which will add the same functionality. #151 Maybe we can join forces so get a complete support?

vklindukh:
> Great contribution @vklindukh. However there is already an open PR which will add the same functionality. #151 Maybe we can join forces so get a complete support?

Looks good to me

PR #180, opened 2020-06-01
Switch to GitHub Actions

crazy-max:
@mc1arke First of all thank you for offering this plugin!

As a gift, here is a PR that offers a switch to GitHub Action for your CI environment to have a minimum of devops maintenance but also something more versatile.

Live example available here: https://github.com/crazy-max/sonarqube-community-branch-plugin/actions/runs/121906900

## Jobs

There are 3 jobs involved in the workflow used.

### Snapshot

This job is in charge of doing a build at each push and also at a pull request. The jar will be stored in the [temporary artifacts space](https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts) which will allow us to quickly test a PR by downloading it:

image

### Release

This job, as the name implies, will create a release using the plugin you currently have in place (`net.researchgate.release`) and publish it on GitHub:

image

The release will be triggered when **you create a new Git tag** for your release (e.g. `1.3.2`). The release on GitHub will be a draft so that you can edit its description before publishing it:

image

And as it already is, the version will be bumped into `gradle.properties`:

image

### Sonar

This job does the analysis of the project on Sonarcloud. You'll just need to [register some GitHub secrets](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets-for-a-repository) on your repo:

* `SONAR_TOKEN`: Token from Sonarcloud
* `SONAR_ORG_KEY`: Organization (should be `mc1arke-github` for you)
* `SONAR_PROJECT_KEY`: Project key (should be `mc1arke_sonarqube-community-branch-plugin` for you)

image


.github/workflows/build.yml
```diff
@@ -0,0 +1,178 @@
...
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      -
+        name: Sonar
+        run: |
+          ./gradlew sonarqube \
+            -Dsonar.host.url=https://sonarcloud.io \
+            -Dsonar.organization=${SONAR_ORG_KEY} \
+            -Dsonar.projectKey=${SONAR_PROJECT_KEY}
```
mc1arke, line 173: I've not tried executing this, but does it not need to specify `-Dsonar.pullrequest.*` or `-Dsonar.branch.*` to prevent reporting of changes against the master branch?

.github/workflows/build.yml
```diff
@@ -0,0 +1,178 @@
+name: build
+
+on:
+  push:
+    branches:
+      - 'master'
+      - 'sq-*'
```
mc1arke, line 7: Any reason for only building on push to these these branches? Whilst the PR job will pick up the build if targeting master or one of the (unreleased) support branches, I'd actually expect to run LTS releases from `release/*` which wouldn't be covered under these patterns. Ideally I'd run the action on all PRs and Pushes, but didn't know if there was a reason you'd applied these restrictions.

.github/workflows/build.yml
crazy-max: @mc1arke SonarQube Gradle plugin takes of it like Maven one or I miss something

.github/workflows/build.yml
crazy-max: @mc1arke Yes you're right we could build for all branches that's not a problem.

.github/workflows/build.yml
mc1arke: Not that I'm aware of, but I'll check. As far as I was aware, it always treats the current branch as the primary branch (normally master) unless you explicitly tell it otherwise.

crazy-max: @mc1arke Ok I will check this out too. We can make some tests [on my fork](https://github.com/crazy-max/sonarqube-community-branch-plugin) if you want, I have enabled Sonarcloud on it: https://sonarcloud.io/dashboard?id=crazy-max_sonarqube-community-branch-plugin

.github/workflows/build.yml
```diff
@@ -0,0 +1,220 @@
...
+              -Dsonar.pullrequest.base=${{ github.event.base_ref }}
+          else
+            ./gradlew sonarqube \
+              -Dsonar.pullrequest.provider=github \
+              -Dsonar.pullrequest.github.repository=${{ github.repository }} \
+              -Dsonar.host.url=https://sonarcloud.io \
+              -Dsonar.organization=${SONAR_ORG_KEY} \
+              -Dsonar.projectKey=${SONAR_PROJECT_KEY} \
+              -Dsonar.branch.name=${GITHUB_REF##*/}
+          fi
```
crazy-max, line 215: @mc1arke I have made some changes to handle Sonarcloud parameters correctly. Now we handle all branches, pull request and releases: https://sonarcloud.io/dashboard?id=crazy-max_sonarqube-community-branch-plugin

image

image

image

mc1arke:

crazy-max:
@mc1arke You're welcome! Looks like it's working great: https://github.com/mc1arke/sonarqube-community-branch-plugin/actions/runs/127541420

PR #182, opened 2020-06-06
Support for bitbucket cloud code insights

marvin-w:
### Intention

This PR intends to provide support for the code insights feature for bitbucket cloud.

### History

Since about 2 months bitbucket cloud also has a code insights feature that one can use. After checking the differences between the cloud and the server implementation it is however not possible to completely reuse the server logic due to renamed/missing fields in the cloud version.

I haven't changed any public facing API (configs, properties, decorators) in this PR which allows @mc1arke to keep it up to date with the other sonarqube versions.

### Additions

* The API is fully functional now for both the server and the cloud version. One development limitation is that Bitbucket actually checks the URL and doesn't allow localhost URLs. That's why someone testing this with a local instance or an instance whose DNS isn't publicly accessible (not confirmed) might have troubles.
* I tried to create a wiki page for this new decorator, however according to https://github.community/t/how-to-fork-a-wiki-and-send-a-pr/2006 it is currently not possible to fork a wiki repository. Is it possible to be granted access to feature branches for the wiki git - then I could create a PR. Alternatively, setting up a dedicated wiki repository could also make sense in case other people are interested in contributing to it.

### Screenshots

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Use the following table to find the correct plugin version for each SonarQube ve
 
 SonarQube Version | Plugin Version
 ------------------|---------------
-7.8 - 8.0         | 1.3.0
+7.8 - 8.0         | 1.3.1
 7.4 - 7.7         | 1.0.2
 
 # Features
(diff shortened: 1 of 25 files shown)


marvin-w:
Due to branching issues I created a new one: #185

PR #183, opened 2020-06-07
Fix Sonarcloud scan for PR

crazy-max:
@mc1arke Ref. https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/58#issuecomment-640186946


mc1arke:
Is it worth making the Sonarqube task dependent on the presence of the `SONAR_TOKEN` secret? Until Github work out an easy way of handling secrets across forks, anyone who submits a PR, or even commits to their own fork without having setup the local secrets is going to get a check failure which isn't a great experience.

mc1arke:
Superseded by #184. Thanks for taking the time to look at this anyway,

PR #184, opened 2020-06-07
Fix execution of Sonarqube on forked repositories and pull requests

mc1arke:
Github does not expose secrets during the execution of actions on pull requests from forked projects, so the Sonar task would always fail on external pull requests. Additionally, anyone forking the project would have had the Sonarqube step fail unless they'd specifically setup Sonarqube secrets on their fork. To overcome this, the `sonar` step is now executed based on the condition of the `SONAR_TOKEN` secret being present in the current execution environment.

Includes a fix for incorrect syntax and variable references in the Pull Request Sonar analysis block.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -184,6 +184,7 @@ jobs:
             ${{ runner.os }}-gradle-
       -
         name: Sonar
+        if: ${{ env.SONAR_TOKEN != null }}
         run: |
           if [[ $GITHUB_REF == refs/tags/* ]]; then
             ./gradlew sonarqube \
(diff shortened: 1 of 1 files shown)


crazy-max:
@mc1arke I think you can authorize the secret to be exposed for a PR with this setting:

image

This works for me on [here](https://github.com/crazy-max/diun/pull/72) for example

mc1arke:
Those settings specify whether you'd allow a third-party action to execute, but don't cover what secrets they're allowed access to. The secrets screen indicates the secret wont be exposed to actions executed on a PR from a fork:

image

PR #186, opened 2020-06-07
Fix invalid base_ref for release job

crazy-max:
@mc1arke Ref. https://github.com/mc1arke/sonarqube-community-branch-plugin/runs/747411860#step:8:75

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
 
   release:
     needs: prep
-    if: needs.prep.outputs.release == 'true'
+    if: github.event_name != 'pull_request' && needs.prep.outputs.release == 'true'
     runs-on: ubuntu-latest
     steps:
       -
@@ -161,16 +161,7 @@ jobs:
         name: Sonar
         if: ${{ env.SONAR_TOKEN != null }}
         run: |
-          if [ "${{ needs.prep.outputs.release }}" = 'true' ]; then
-            ./gradlew sonarqube \
-              -Dsonar.pullrequest.provider=github \
-              -Dsonar.pullrequest.github.repository=${{ github.repository }} \
-              -Dsonar.host.url=https://sonarcloud.io \
-              -Dsonar.organization=${SONAR_ORG_KEY} \
-              -Dsonar.projectKey=${SONAR_PROJECT_KEY} \
-              -Dsonar.projectVersion=${{ needs.prep.outputs.version }}
-          elif [ "${{ github.event_name }}" = "pull_request" ]; then
-            PR_BRANCH=${{ github.head_ref }}
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
             ./gradlew sonarqube \
               -Dsonar.pullrequest.provider=github \
               -Dsonar.pullrequest.github.repository=${{ github.repository }} \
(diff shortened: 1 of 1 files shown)


crazy-max:
@mc1arke 

image

image

image

.github/workflows/build.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         name: Checkout
         uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.base_ref }}
+          ref: ${{ github.event.repository.master_branch }}
```
mc1arke, line 74: This is going to prevent releasing off the `release/*` branches which will cause a longer-term issue with back-porting features/fixes to the version that supports the LTS Sonarqube version. Given the priority just now is to release from master so I can get the 8.1 and 8.2 supporting versions out, this isn't a critical issue, but do you have any suggestions for allowing non-master releases?

crazy-max: @mc1arke We can take a look at the [Dump context](https://github.com/mc1arke/sonarqube-community-branch-plugin/runs/747411860#step:3:184) to correlate with the information we need. I will make some tests on my fork.

crazy-max: @mc1arke From what I could find and confirm is that all the projects I could find tagged their release only from the master branch.

Looking at the best practices, the Git flow is based on [this process](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) for the release branches:

> Once the release is ready to ship, it will get merged it into master and develop, then the release branch will be deleted.

.svg?cdnVersion=1060)

The same applied for this [Git branching model](https://nvie.com/posts/a-successful-git-branching-model/#release-branches):

```
$ git checkout master
Switched to branch 'master'
$ git merge --no-ff release-1.2
Merge made by recursive.
(Summary of changes)
$ git tag -a 1.2
```

mc1arke: It wouldn't be practical to try and release this plugin only from master. Since Sonarqube make binary/functional incompatible changes between some of their releases and I'm generally trying to maintain a support the current Sonarqube long-term support version, I need a way of releasing off non-master.

I'm aware this doesn't make releasing particularly simple to manage through an action, so am happy to continue manual releases and have the release job removed from the actions if required.

crazy-max: > It wouldn't be practical to try and release this plugin only from master. Since Sonarqube make binary/functional incompatible changes between some of their releases and I'm generally trying to maintain a support the current Sonarqube long-term support version, I need a way of releasing off non-master.

Indeed if you are working on several releases for backwards compatibility, there is no need to create releases only from the master branch.

> I'm aware this doesn't make releasing particularly simple to manage through an action, so am happy to continue manual releases and have the release job removed from the actions if required.

The only problem here is that you have to push a file on the reference branch to the tag at each release. For me this kind of file "should" be immutable and can be summed up to `version=dev`.
But we can always find the reference to the tag and do a proper checkout at the time of release. I'll keep you in touch.

crazy-max: @mc1arke Ok I've made some changes do deal with branch ref for a tag.

https://github.com/crazy-max/sonarqube-community-branch-plugin/runs/748014873?check_suite_focus=true

image

.github/workflows/build.yml
```diff
@@ -71,19 +71,29 @@ jobs:
...
         uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.base_ref }}
+          fetch-depth: 0
+      -
+        name: Prepare
+        id: prepare
+        run: |
+          echo ::set-output name=tag_name::${GITHUB_REF#refs/tags/}
+          echo ::set-output name=branch_name::$(git --no-pager branch -a --contains "$GITHUB_SHA" | grep 'remotes/origin/' | cut -d '/' -f3)
```
mc1arke, line 80: The `cut` section is going to incorrectly truncate the branch name if it contains a slash. If I use the branch `release/1_3` as an example, this would find the commit on that branch, but then truncate it to `release` which wont match any available branch

mc1arke, line 80: This seems like it has the potential of choosing the wrong branch if a commit appears on multiple branches, such as we have just now on the `release/1_3` branch in this repo

crazy-max:
@mc1arke Ok now I check the gradle props through and if it's a non semver I use the value as our base ref.

crazy-max:
@mc1arke

> check if the properties file was non-snapshot and would execute the release plugin if a non-snapshot version was found.

Ok I get it!

Release workflow: https://github.com/crazy-max/sonarqube-community-branch-plugin/runs/798797932?check_suite_focus=true
Release published: https://github.com/crazy-max/sonarqube-community-branch-plugin/releases/tag/1.3.3

PR #193, opened 2020-06-13
BitBucket decoration fix

vklindukh:
Below are couple of fixes of BitBucket decoration in sq-8_2-support branch

- replace .getAlmRepo() and getAlmSlug() calls (to get correct parameters order in "insights/1.0/projects/%s/repos/%s/" templates)
- fix almSettingDto.getUrl() call (missed method)
- add debug logging (well, for debug)


mc1arke:
Since these changes were part of the work for Sonarqube 8.1, could you update your commit and PR to point at master please? That way they'll be automatically picked up for the 8.2 release.

PR #199, opened 2020-06-25
Bump SonarQube version

vklindukh:
Bump SonarQube version to the latest 8.3.1.34397

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ repositories {
     }
 }
 
-def sonarqubeVersion = '8.2.0.32929'
+def sonarqubeVersion = '8.3.1.34397'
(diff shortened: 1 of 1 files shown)


mc1arke:
Why are we wanting to bump the base Sonarqube version? This would make the 8_2 branch unable to run against Sonarqube 8.2 which doesn't seem right.

vklindukh:
> Why are we wanting to bump the base Sonarqube version? This would make the 8_2 branch unable to run against Sonarqube 8.2 which doesn't seem right.

SQ 8.3.1 is the current version, this is the only reason

nixel2007:
But there is no need to bump version to support plugin in 8.3...

PR #213, opened 2020-07-20
[GITHUB] Add support for multiple pull request decoration - mono repo support

useful-devops-tools:
We have a mono repo with multiple microservices in it.
If multiple microservices are modified at once then only the final sonarqube PR decoration appeared in the checks tab of Github. This change updates the name of the Status check to include the sonarqube project name so that multiple checks can be attached to the one pull request.

Tested and verified the functionality on Sonarqube 8.3


mc1arke:

Issue #135, opened 2020-03-22
[Gitlab] Filtering by severity which issues are added to MR


bmaehr:
**Describe the Enhancement**
Filtering by severity which issues are added to MR

**Additional context**
Because it is not possible to disable many rules on sonar we put the ones we don't want to level info and don't want them to be as comment on the MR.

4n4n4s:
Hi @bmaehr 
should be pretty easy to implement to add a severity threshold.
Besides that adding a total issue threshold would be good so that not thousands of issues are pushed to bitbucket in case something goes terribly wrong.

Are you going to implement it?

@mc1arke  What's your opinon?

mc1arke:
Limiting the number of reported issues sounds reasonable, providing it's then clear in a summary or comment that the output is truncated.

Does filtering not still cause an issue with the quality gate failing but issues not being decorated?

bmaehr:
I have reimplemented the change based on master if someone is interested ...

bmaehr:
@mc1arke I'm still buliding my own version of your project containing for example this feature and would like to try again to add it to the master.

PR #214, opened 2020-07-20
Bugfix: #201 could not decorate gitlab merge request

PiekJ:
Enables GitLab merge request to be decorated. This PR fixes issue #201 for SonarQube 8.2+.

Issue #96, opened 2020-02-20
disable file comment if the number of issues in a pr excceed some  value


jizhilong:
with file comment enabled,  the pr decoration gate will send several comments to gitlab/github/bitbucket. Which is really nice in most situation, but if someone accidentally pushed 
a bunch of tool generated code to a merge request, the number of issues will be horrible,  posting file comments in such a situation will not help.

PR #217, opened 2020-07-20
Feature: #135 #96 comments filter and limit

PiekJ:
Adds functionality for decorator comments to filter by severity, type and limit the amount of comments. This PR fixes issue #135 and #96. This PR depends on PR #214.


PiekJ:
Removed PR #214 dependency so it can be reviewed independently. Squashed the commits too.

PR #224, opened 2020-08-06
fix code smell image path in the analysis summary

dznaiko:
I have noticed that the Code Smell group in the PR decoration uses the same icon as the Vulnerabilities group.
This PR fixes this minor bug. It also includes the updated tests.


mc1arke:

PR #226, opened 2020-08-07
Resolve summary comment on GitLab MRs when the Quality Gate has passed

teake:
MRs where the Quality Gate passes do not a priori need additional action.
Hence the summary comment can be resolved automatically by SonarQube.


yzhoholiev:
Hi! We have a similar modification in our fork. It would be great to make it part of the plugin.

mc1arke (changes requested):
Hi @teake.

Thanks for the contribution and sorry for the delayed review. Could you add a test to ensure your new code is covered please? I need to ensure that any code can be quickly verified during any future changes without me having to exhaustively test multiple manual combinations.

Thanks!

teake:
@mc1arke Done -- at least, I suspect so. I haven't run locally with coverage to check if it now is more than 80% but I guess this should do the trick.

adpeyre:
@mc1arke Is it possible to merge this pull request. 
The behavior is annoying because Sonarqube comment is deleted but the thread is still present.

image

mc1arke (approved):

mc1arke:
@adpeyre your issue is something different: you're seeing threads created from inline code-comments where an entry still exists where the thread used to be, whereas this is the summary comment thread being changed here which isn't an in-line comment.

mc1arke:

PR #242, opened 2020-09-14
Add support for Drone CI autoconfiguration from environment variables

lafriks:
Drone CI environment variable reference: <https://docs.drone.io/pipeline/environment/reference/>


mc1arke:
Thanks for the proposal, but I've decided that I'm not currently going to accept this contribution. I'm keeping the feature-set in this plugin roughly in parity with the features of the official Sonarqube distribution which doesn't support Drone CI, and I don't particularly want to have the branch detection growing with multiple possible environment detections.

PR #258, opened 2020-10-16
Reduce memory usage of the PostAnalysisIssueVisitor

thomasgl-orange:
The collection of `DefaultIssue` objects hold by the `PostAnalysisIssueVisitor` can take a significant amount of memory in the CE process heap, when there are numerous issues (a `DefaultIssue` alone can weigh several 10th of kilobytes). 

A small example in Eclipse MAT (and I've seen much worse):
heap

This PR mostly fixes that. The logic is unchanged: there is still a list of issues to be processed at the end. But it contains lighter objects, with only the few members which would actually be used later to decorate pullrequests. On our production instances (1.3.2 plugin / SQ 7.9 LTS), a similar patch made a major difference on the CE processes memory footprint.

I've tried to write this patch to minimize impact/changes on the code. The `PostAnalysisIssueVisitor.LightIssue` class exposes a subset of the `DefaultIssue` public interface, thus the changes are mostly limited to the actual return type of `PostAnalysisIssueVisitor.ComponentIssue#getIssue()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,11 @@
 /.idea/
 *.iml
 
+# Eclipse
+/.classpath
+/.project
+/.settings/
+/bin/
+
 #Project libs
 /sonarqube-lib/
--- a/build.gradle
+++ b/build.gradle
@@ -68,6 +68,7 @@ dependencies {
     compile('io.aexp.nodes.graphql:nodes:0.5.0') {
         exclude group: 'com.fasterxml.jackson.core'
     }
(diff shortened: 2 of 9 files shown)


mc1arke:
Thanks for the contribution. Please fix the issues highlighted by Sonarqube and ensure your new code is covered.

thomasgl-orange:
@mc1arke, I've removed the useless import, and added unit tests for the `PostAnalysisIssueVisitor.LightIssue` bean definition.

mc1arke:

Issue #255, opened 2020-10-15
SonarQube 8.5+ support


augustomezencio-hotmart:
Hello!

We have used this plugin in the 1.3.2 version with SonarQube 7.8 and it works really great.
After updating our SonarQube to 8.5 version and the plugin for 1.4.0 we are receiving 500 error when it try to upload the reports.
Do you have plans to update the support for the SonarQube 8.5+ ?

Regards.

nixel2007:
1.4.0 does not support anything rather than 8.1. You should try to build jar from master

brunomendola:
Hi, this is the related stacktrace I see in web.log

```
2020.10.16 13:14:28 ERROR web[][o.s.s.p.w.RootFilter] Processing of request /api/ce/submit?projectKey=project1&projectName=Project1&characteristic=branch%3Dmaster&characteristic=branchType%3DBRANCH failed
javax.servlet.ServletException: Filter execution threw an exception
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:200)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.sonar.server.platform.web.UserSessionFilter.doFilter(UserSessionFilter.java:80)
        at org.sonar.server.platform.web.UserSessionFilter.doFilter(UserSessionFilter.java:67)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.sonar.server.platform.web.CacheControlFilter.doFilter(CacheControlFilter.java:76)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
        at org.sonar.server.platform.web.SecurityServletFilter.doHttpFilter(SecurityServletFilter.java:76)
        at org.sonar.server.platform.web.SecurityServletFilter.doFilter(SecurityServletFilter.java:48)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
[...]

PR #259, opened 2020-10-16
#255 Compatibility with SonarQube CE 8.5.0.37579

brunomendola:
SonarQube introduced breaking changes in `org.sonar.server.ce.queue.BranchSupport`, so the plugin had to be adapted.

This change breaks backward compatibility, meaning that the plugin will now only support SonarQube 8.5.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/server/CommunityComponentKey.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/server/CommunityComponentKey.java
@@ -18,6 +18,7 @@
  */
 package com.github.mc1arke.sonarqube.plugin.server;
 
+import org.sonar.db.component.BranchType;
 import org.sonar.server.ce.queue.BranchSupport;
 
 import java.util.Optional;
@@ -29,13 +30,15 @@ import java.util.Optional;
 
     private final String key;
     private final String dbKey;
-    private final BranchSupport.Branch branch;
+    private final String branchName;
+    private final BranchType branchType;
     private final String pullRequestKey;
 
-    /*package*/ CommunityComponentKey(String key, String dbKey, BranchSupport.Branch branch, String pullRequestKey) {
+    /*package*/ CommunityComponentKey(String key, String dbKey, String branchName, BranchType branchType, String pullRequestKey) {
         this.key = key;
         this.dbKey = dbKey;
-        this.branch = branch;
+        this.branchName = branchName;
+        this.branchType = branchType;
         this.pullRequestKey = pullRequestKey;
     }
 
@@ -50,8 +53,12 @@ import java.util.Optional;
     }
 
     @Override
-    public Optional<BranchSupport.Branch> getBranch() {
-        return Optional.ofNullable(branch);
+    public Optional<String> getBranchName() {
+        return Optional.ofNullable(branchName);
+    }
+
+    public Optional<BranchType> getBranchType() {
+        return Optional.ofNullable(branchType);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


src/test/java/com/github/mc1arke/sonarqube/plugin/classloader/ClassReferenceElevatedClassLoaderFactoryTest.java
```diff
@@ -45,6 +46,8 @@
 /**
  * @author Michael Clarke
  */
+@Ignore("SonarQube has removed bundled plugins from extensions/plugins.\n" +
+    "See https://jira.sonarsource.com/browse/SONAR-13792")
```
mc1arke, line 50: I'm not happy with these tests being ignored:
1. The classloader functionality is key to how the plugin works, so needs to have something that proves it works during any changes
2. Marking a test as ignored is likely to lead to it never being maintained; if the test isn't needed then we should delete it

I suspect that these tests can continue to operate if the target class is modified, the build classpath is updated to remove the new extensions directory, and the classloader initialisation in the test updated to reference the new extensions directory.

src/main/java/com/github/mc1arke/sonarqube/plugin/server/CommunityComponentKey.java
```diff
@@ -29,13 +30,15 @@
 
     private final String key;
     private final String dbKey;
-    private final BranchSupport.Branch branch;
+    private final String branchName;
+    private final BranchType branchType;
     private final String pullRequestKey;
 
-    /*package*/ CommunityComponentKey(String key, String dbKey, BranchSupport.Branch branch, String pullRequestKey) {
+    /*package*/ CommunityComponentKey(String key, String dbKey, String branchName, BranchType branchType, String pullRequestKey) {
```
mc1arke, line 37: The `BranchType` is probably unnecessary here. `ReportSubmitter` in Sonarqube core will only call the `Community ranchSupportDelegate#createBranchComponent` method if the branch does not already exist, and will be passing in the 'main' branch as the component DTO for comparison, and for basing any new component on. The branch type would only therefore make a difference if the new component matched the name of the main branch but was a pull request, but this is specifically defended against in `ReportSubmitter`

mc1arke (changes requested):
As per my comments, I'd like the tests not to be ignored

src/main/java/com/github/mc1arke/sonarqube/plugin/server/CommunityComponentKey.java
brunomendola: There was a specific condition at `CommunityBranchSuppirtDelegate:93` that I maintained by passing the `branchType` arg. Do you think that condition is not necessary anymore because of some changes in 8.5 or it already was unnecessary? If it is so, we can simplify a lot.

src/test/java/com/github/mc1arke/sonarqube/plugin/classloader/ClassReferenceElevatedClassLoaderFactoryTest.java
brunomendola: The problem here is that SonarQube doesn't ship any extension in the release zip anymore. They choose to embed the extensions that were there in the previous versions, so the extensions directory is now empty.
What we can do, if you want to keep the tests, is make Gradle download a specific extension jar into the extensions directory only for tests. What do you think?

mc1arke:
I've raised https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/262/files with what I believe are the minimum changes required to get 8.5.0 working.

brunomendola:
Ok man, no problems at all. 
Thanks for clarifications and for your great work on this plugin.

PR #262, opened 2020-10-25
Add support for Sonarqube 8.5.0

mc1arke:
Sonarqube 8.5.0 removed the `BranchSupport.Branch` class which breaks backwards-compatibility with previous Sonarqube versions, and requires updating how the ComponentKey is checked and generated during the creation of the analysis report. This version of Sonarqube also moves the bundled plugins into their own directory, so the build classpath has been updated to exclude those plugins from being able to be part of the compilation path, and updates the class-loading unit tests to reference a plugin class and file path that is still available in the newer Sonarqube version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ repositories {
     }
 }
 
-def sonarqubeVersion = '8.2.0.32929'
+def sonarqubeVersion = '8.5.0.37579'
 def sonarqubeLibDir = "${projectDir}/sonarqube-lib"
 def sonarLibraries = "${sonarqubeLibDir}/sonarqube-${sonarqubeVersion}/lib"
 
@@ -57,8 +57,8 @@ compileJava {
 
 dependencies {
     testCompile group: 'junit', name: 'junit', version: '4.12'
-    compileOnly fileTree(dir: sonarLibraries, include: '**/*.jar')
-    testCompile fileTree(dir: sonarLibraries, include: '**/*.jar')
+    compileOnly fileTree(dir: sonarLibraries, include: '**/*.jar', exclude: 'extensions/*.jar')
+    testCompile fileTree(dir: sonarLibraries, include: '**/*.jar', exclude: 'extensions/*.jar')
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'org.mockito', name: 'mockito-core', version: '3.1.0'
     testCompile group: 'org.assertj', name: 'assertj-core', version: '3.13.2'
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/server/CommunityBranchSupportDelegate.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/server/CommunityBranchSupportDelegate.java
@@ -71,9 +71,8 @@ public class CommunityBranchSupportDelegate implements BranchSupportDelegate {
         String branch = StringUtils.trimToNull(characteristics.get(CeTaskCharacteristicDto.BRANCH_KEY));
 
         try {
-            BranchType branchType = BranchType.valueOf(branchTypeParam);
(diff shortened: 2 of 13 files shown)


rpo-fr:
Hi,
when i merge this branch on master, then build with gradle clean build, my build failed

`
> Task :test

com.github.mc1arke.sonarqube.plugin.classloader.ReflectiveElevatedClassLoaderFactoryTest > testLoadClassInvalidApiClassloader FAILED
    java.lang.NullPointerException at ReflectiveElevatedClassLoaderFactoryTest.java:150

com.github.mc1arke.sonarqube.plugin.classloader.ReflectiveElevatedClassLoaderFactoryTest > testLoadClassInvalidClassRealmKey FAILED
    java.lang.NullPointerException at ReflectiveElevatedClassLoaderFactoryTest.java:96

com.github.mc1arke.sonarqube.plugin.classloader.ReflectiveElevatedClassLoaderFactoryTest > testLoadClassNoParentRef FAILED
    java.lang.NullPointerException at ReflectiveElevatedClassLoaderFactoryTest.java:123

com.github.mc1arke.sonarqube.plugin.classloader.ReflectiveElevatedClassLoaderFactoryTest > testLoadClassInvalidClassloader FAILED
    java.lang.NullPointerException at ReflectiveElevatedClassLoaderFactoryTest.java:175

com.github.mc1arke.sonarqube.plugin.classloader.ReflectiveElevatedClassLoaderFactoryTest > testLoadClass FAILED
    java.lang.NullPointerException at ReflectiveElevatedClassLoaderFactoryTest.java:66

com.github.mc1arke.sonarqube.plugin.classloader.ClassReferenceElevatedClassLoaderFactoryTest > testClassloaderReturnedOnHappyPath FAILED
    java.lang.NullPointerException at ClassReferenceElevatedClassLoaderFactoryTest.java:112

com.github.mc1arke.sonarqube.plugin.classloader.ClassReferenceElevatedClassLoaderFactoryTest > testLoadClass FAILED
    java.lang.NullPointerException at ClassReferenceElevatedClassLoaderFactoryTest.java:112

283 tests completed, 7 failed
`

Do you have the same?

mc1arke:
@rpo-fr you'll need to ensure you've deleted any old versions of Sonarqube from the sonarqube-lib directory. The unit tests currently use the first directory they find in there as the target for testing against, so are likely to be using an older version of Sonarqube that doesn't have the extensions/plugins directory in it, and are therefore failing when trying to list the files in there

enginates:
on sonarqube-8.5.0.37579 
I am getting this error
java.lang.IllegalStateException: Current edition does not support branch feature

nixel2007:
Is it normal that I can't save GitHub Client ID and GitHub Client Secret at ALM settings on SQ 8.5.1? GitHub API URL, App ID and Private key are saved without problems, bit client id and client secret cleared after page reload

mc1arke:
@nixel2007 The plugin doesn't (currently) use those fields so doesn't save them. They were introduced by SONAR-13524 in 8.4 of Sonarqube but they've not been added to this plugin's implementation of the relevant API yet. I'll add them in, and have a look at whether we should be using those fields as part of our Github authentication or API calls.

cambierr:
I may have missed something but when installing this plugin on by 8.5.1 Sonarqube, I don't see any control that I had before while it's visible in the list of installed plugins. Any idea ?

enginates:
> I may have missed something but when installing this plugin on by 8.5.1 Sonarqube, I don't see any control that I had before while it's visible in the list of installed plugins. Any idea ?

Did you follow this?
In case you want to try and test the current branch or build it for your development execute ./gradlew clean build inside of the project directory. This will put the built jar under libs/sonarqube-community-branch-plugin*.jar

cambierr:
Thank you @enginates I recently switched to a helm-based deploy instead of a manual one and I always forgot that lib/common folder... Thank you very much... will test this :)

bameur:
Do you have a date for the next release (sonar 8.5)?
Thank you very much.

PR #269, opened 2020-11-09
Remove the duplicate JUnit dependency from build.gradle

dznaiko:
I noticed a duplicate dependency and thought that the project could do without it :)


mc1arke:

Issue #295, opened 2021-01-12
[SQ8.6] plugin missing API path /api/alm_settings/validate


williamm-ic:
**Describe the bug**
Starting with SonarQube 8.6 we have a new page for the ALM integrations. It now requires a name, collection URL and PAT (whereas v8.5 and lower only require a PAT). The configuration is validated once created and an error message appears ("Unknown url : /api/alm_settings/validate"). 

**To Reproduce**

_For fresh installations:_
1. Go to 'Administration > General Settings > ALM Integrations > Azure DevOps Server'
2. Click on 'Create Configuration'
3. Fill in the form and save it
4. See error (on top of the screen or with DevTools)

_For existing configurations:_
1. Go to 'Administration > General Settings > ALM Integrations > Azure DevOps Server'
2. Click on 'Check configuration'
3. See error (on top of the screen or with DevTools)

**Expected behavior**
The configuration is validated and, if successful, can be used for Pull Request Decoration and Project Import.

**Screenshots**
-

**Software Versions**
 - SonarQube Version: 8.6.0 (build 39681)
 - Plugin Version: 1.6.0

**Additional context**
-

djepoulet:
Hello There, 

It seems the 8.4 is not fully compatible, i've the following message when i'm trying to add the GitLab ALM.. 

"Could not validate this configuration."
ERROR: Unknown url : /api/alm_settings/validate****

Do someone have fully tested 8.4? Is it compatible with the last version of the plugin?

williamm-ic:
> Do someone have fully tested 8.4? Is it compatible with the last version of the plugin?

AFAIK it's not a 100% compatible. 
Not only Azure is affected, but we also have a confirmation for GitLab. How about BitBucket Server?

djepoulet:
¨Perhaps in this case, it's pretty obvious but, should we not edit the release and specify that the lastest version: 

"...supports Sonarqube 8.5 and above"

But : 

"...supports Sonarqube 8.5."

:)

HaddadJoe:
AML integration for github is also facing the same issue
--
Version: 8.6.0.39681
Plugin Version: 1.6.0

123Haynes:
Can you guys try the attached jar? I've updated the sonarqube version in it which should fix the issue I think.

[...]

PR #300, opened 2021-01-30
Docs: Update compability matrix with sonar 8.6

123Haynes:
I tested v1.6.0 with sonar 8.6.1.40680 at our company installation and it works without any problems.

I also updated the sonar version in build.gradle and all tests still worked fine after that.
If you want I can also add that change to this PR, but since the already released version works just fine it would be better keep codechanges separate from this. 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Use the following table to find the correct plugin version for each SonarQube ve
 
 SonarQube Version | Plugin Version
 ------------------|---------------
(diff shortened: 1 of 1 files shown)


123Haynes:
I just saw https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/295 We don't use that integration so we didn't have that problem. Closing this as it is not 100% compatible.

PR #308, opened 2021-02-25
Fix GitLab Status URL

keyiyin:
Hi,
I find post GitLab MR pipeline status does not work, because GitLab MR pipeline is from source project , but not from target project.


mc1arke (approved):

mc1arke:

PR #312, opened 2021-03-08
Add ability to generate Docker Image containing plugin

mc1arke:
This change adds the relevant docker files to allow building a development docker image containing the code in this repository, or generating an image using a released version of the plugin, with either of these being backed by a particular base Docker image of Sonarqube's community releases. The `.env` file contains the variables required to configure what builds is performed with docker-compose.yml then driving the relevant creation from there.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM openjdk:11-jdk-slim as builder
+
+COPY . /home/build/project
+WORKDIR /home/build/project
+RUN ./gradlew build -x test
+
+FROM sonarqube:community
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM openjdk:11-jdk-slim as builder
+
+COPY . /home/build/project
+WORKDIR /home/build/project
+RUN ./gradlew build -x test
+
+FROM sonarqube:community
```
tisoft, line 7: Shouldn't we allow overriding the sonarqube version here, like in the release docker image?

```
ARG SONARQUBE_VERSION

FROM sonarqube:${SONARQUBE_VERSION}

```

The docker-compose.yml already sets the SONARQUBE_VERSION arg, so would support this change.

PR #323, opened 2021-04-03
Hide outdated summary comments on GitHub

VanRoy:
This pull request enhance the support of the summary comment for GitHub by minimizing previous ( outdated ) comments on Pull Request when new commit is pushed on same branch and the analysis is changed.

Screenshot of comment minimized after re-analysis done by SonarQube :

sonar_github_outdated


VanRoy:
Hi @mc1arke , do you think that this enhancement is useful for the project ?
We use it day to day on and it works well for us.

Do not hesitate to tell me if you think it require some changes , more tests or something else.

Thanks a lot.

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GetPullRequest.java
```diff
@@ -49,13 +51,81 @@ public PullRequest getPullRequest() {
 
         private final String id;
 
+        @GraphQLProperty(name = "comments", arguments = {@GraphQLArgument(name = "last", value = "100", type = "Integer")})
```
mc1arke, line 54: The default of `100` could cause problems, as an MR with a large number of comments would only have the last 100 checked. This should really be controlled from the calling code, with an additional field of `totalCount` on the response so we can iterate from `start: 100` to `start: totalCount` in batches of 100 and ensure we've covered all messages

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java
```diff
@@ -222,6 +231,42 @@ private void postSummaryComment(String apiUrl, Map<String, String> headers, Stri
 
     }
 
+    private void minimizeComment(String graphqlUrl, Map<String, String> headers, String commentId) {
+
+        try {
```
mc1arke, line 236: Nitpick: the IOException can only happen on the `executeRequest` as far as I can tell, so keep the try block constrained to that line rather than covering the whole method.

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GraphqlCheckRunProvider.java
```diff
@@ -178,29 +180,36 @@ public DecorationResult createCheckRun(AnalysisDetails analysisDetails, AlmSetti
 
     }
 
-    private void postSummaryComment(String apiUrl, Map<String, String> headers, String projectPath, String pullRequestKey, String summary) throws IOException {
+    void postSummaryComment(String graphqlUrl, Map<String, String> headers, String projectPath, String pullRequestKey, String summary) throws IOException {
```
mc1arke, line 183: If this has been made visible purely for testing, then it indicate this class has potentially got too complex. Is it worth moving this into a `SummaryCommentHandler` class?

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/v4/GetPullRequest.java
```diff
@@ -49,13 +51,81 @@ public PullRequest getPullRequest() {
...
+        public String getId() {
+            return id;
+        }
+
+        public Comments getComments() {
+            return comments;
+        }
+    }
+
+    public static class Comments {
```
mc1arke, line 72: Nitpick: I'm not sure these classes need to be static nested classes of `GetPullRequest`. They are part of the response from GetPullRequest, but would also be valid for anything else that uses Comment nodes

mc1arke (changes requested):
Thanks for the contribution. A few small comments, but I'm happy with the concept of the proposal!

antiqe:
Hi @VanRoy, really nice enhancement 👍 , i think it's something that really missing right now.

mc1arke:
Merged under #435 with a few minor changes. Thanks for the contribution!

Issue #335, opened 2021-04-12
Gitlab PR decoration don't work anymore


Glosur:
**Describe the bug**
Gitlab PR decoration is failing

**To Reproduce**
Configure gitlab PR decoration
**Expected behavior**
No ERROR/failing

**Software Versions**
 - SonarQube Version: [8.7.0] (reproduced on 8.8.0)
 - Plugin Version: 1.7.0

**Additional context**

```

2021.04.12 08:45:23 ERROR ce[AXjFQa7dboEK8z4slOrP][o.s.c.t.p.a.p.PostProjectAnalysisTasksExecutor] Execution of task class com.github.mc1arke.sonarqube.plugin.ce.pullrequest.PullRequestPostAnalysisTask failed
java.lang.IllegalArgumentException: Cannot format given Object as a Number
	at java.base/java.text.DecimalFormat.format(Unknown Source)
	at java.base/java.text.Format.format(Unknown Source)
	at com.github.mc1arke.sonarqube.plugin.ce.pullrequest.AnalysisDetails.createAnalysisSummary(AnalysisDetails.java:207)
	at com.github.mc1arke.sonarqube.plugin.ce.pullrequest.gitlab.GitlabServerPullRequestDecorator.decorateQualityGateStatus(GitlabServerPullRequestDecorator.java:171)
	at com.github.mc1arke.sonarqube.plugin.ce.pullrequest.PullRequestPostAnalysisTask.finished(PullRequestPostAnalysisTask.java:160)
	at org.sonar.ce.task.projectanalysis.api.posttask.PostProjectAnalysisTasksExecutor.executeTask(PostProjectAnalysisTasksExecutor.java:118)
	at org.sonar.ce.task.projectanalysis.api.posttask.PostProjectAnalysisTasksExecutor.finished(PostProjectAnalysisTasksExecutor.java:109)
	at org.sonar.ce.task.step.ComputationStepExecutor.executeListener(ComputationStepExecutor.java:91)
	at org.sonar.ce.task.step.ComputationStepExecutor.execute(ComputationStepExecutor.java:63)
	at org.sonar.ce.task.projectanalysis.taskprocessor.ReportTaskProcessor.process(ReportTaskProcessor.java:81)
	at org.sonar.ce.taskprocessor.CeWorkerImpl$ExecuteTask.executeTask(CeWorkerImpl.java:235)
[...]

Issue #325, opened 2021-04-06
Gitlab MR decoration failed (Cannot format given Object as a Number)


Atharex:
**Describe the bug**
The plugin enables importing Gitlab projects and scanning pull requests correctly, but the MR decoration fails.

**To Reproduce**

```
sonar-scanner   -Dsonar.projectKey=gitlab_project_AXinYBuZsjqXafDIc7B1   -Dsonar.sources=.   -Dsonar.host.url=<sonarqube-url>   -Dsonar.login=<token> -Dsonar.pullrequest.key=4 -Dsonar.pullrequest.branch=feature/458 -Dsonar.pullrequest.base=master
```

This correctly analyzes the branch and compares it to master. However sonarqube logs show that decoration has failed ("Cannot format given Object as a Number") with the following output:

```
...
2021.04.06 13:51:37 INFO  ce[AXinc-8ksjqXafDIc7CN][o.s.c.t.p.a.p.PostProjectAnalysisTasksExecutor] Webhooks | globalWebhooks=0 | projectWebhooks=0 | status=SUCCESS | time=3ms
2021.04.06 13:51:37 INFO  ce[AXinc-8ksjqXafDIc7CN][c.g.m.s.p.c.p.PullRequestPostAnalysisTask] using pull request decorator com.github.mc1arke.sonarqube.plugin.ce.pullrequest.gitlab.GitlabServerPullRequestDecorator
2021.04.06 13:51:37 INFO  ce[AXinc-8ksjqXafDIc7CN][c.g.m.s.p.c.p.g.GitlabServerPullRequestDecorator] starting to analyze with com.github.mc1arke.sonarqube.plugin.ce.pullrequest.AnalysisDetails@631b8200
2021.04.06 13:51:37 INFO  ce[AXinc-8ksjqXafDIc7CN][c.g.m.s.p.c.p.g.GitlabServerPullRequestDecorator] class com.github.mc1arke.sonarqube.plugin.ce.pullrequest.gitlab.response.MergeRequest received
2021.04.06 13:51:37 INFO  ce[AXinc-8ksjqXafDIc7CN][c.g.m.s.p.c.p.g.GitlabServerPullRequestDecorator] Status url is: https://gitlab.com/api/v4//projects/<project-id>/statuses/3c3bd88b77c09cb46dc97450ae2f52c0b46c5173
[...]

PR #343, opened 2021-04-21
Fix NumberFormat call on a non Number (#325, #335)

Rakdos8:
Probably not the best fix but at least it unlocks the MR decoration.


src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/AnalysisDetails.java
```diff
@@ -204,7 +204,7 @@ public String createAnalysisSummary(FormatterFactory formatterFactory) {
                                                          Optional.ofNullable(newCoverage).map(decimalFormat::format)
                                                                  .map(i -> i + "% Coverage")
                                                                  .orElse("No coverage information") + " (" +
-                                                         decimalFormat.format(coverage) + "% Estimated after merge)")),
```
mc1arke, line 207: Nitpick: Avoid using negation in ternaries, as it makes the logic difficult to follow. I'd probably switch this to `Optional.ofNullable(coverage).orElse(0)` which does incur the overhead of an extra object on the heap, but helps makes the code more readable.

Issue #366, opened 2021-06-04
[Gitlab] Per-project configuration for gitlab API token for MR decoration comments gone in 8.9 LTS / 8.x


nejch:
First, thanks a lot for maintaining compatibility with several versions, it must be a pain.

We're in the process of migrating to 8.9 LTS from 7.9 LTS (plugin 1.3.x to 1.8.0) and noticed that, with the new ALM Settings replacing some of the the old Pull Request config, the ability to configure a PAT for PR decoration per-project is gone.

I understand this basically follows the SonarQube feature set and it's convenient for some as it's less config. But when SQ integrates with a large GitLab instance I'm afraid people won't want to share the same token with read access to hundreds of projects that want to integrate with the SQ instance. Is there any chance that the per-project token config option from older versions can be restored? Thanks again!

ezintz:
Not sure if I understand your issue. You can create in SonarQube tokens for different users having access to different projects. You just need to ensure the correct token is used within the GitLab CI job. Is this a possible solution for your request?

nixel2007:
@ezintz it's not about sonarqube tokens in gitlab, it's about gitlab tokens in sonarqube :)

nejch:
> 
> 
> Not sure if I understand your issue. You can create in SonarQube tokens for different users having access to different projects. You just need to ensure the correct token is used within the GitLab CI job. Is this a possible solution for your request?

@ezintz sorry I guess I should have been clearer on that. In previous versions you were able to configure the GitLab API token (usually a GitLab project access token of a bot user) that posts the MR decorations to GitLab, in every project. This section under Pull Request in the Project Administration menu:

![image](https://user-images.githubusercontent.com/16777978/120832593-2056ef80-c561-11eb-9533-6ebbde5a0744.png)

This is now replaced with:

![image](https://user-images.githubusercontent.com/16777978/120832768-4f6d6100-c561-11eb-8bd7-fc578fd364ac.png)

This means only one bot user can be set for the entire SonarQube instance. If people want PR decorations in their GitLab projects they need to give reporter access to that user, so you end up with a token that will have read access to a bunch of projects.

nejch:
Thanks a lot for the quick PR there @mc1arke!

Just wondering as I'm not really knowledgeable in Java, the related PR enables multiple ALM entries but they'd still need to be configured at instance level, or does this bring back the project-level config already?

[...]

PR #367, opened 2021-06-05
#366: Allow defining multiple configurations per ALM

mc1arke:
The default distribution of Sonarqube only allows a single instance of each ALM to be configured, which prevents using ALM decoration where 2 different repositories require different API keys in a single ALM. To allow multiple configurations to be defined per ALM, the `MultipleAlmFeatureProvider` is being intercepted by the agent for the Web component and the `enabled` method being replace to always return `true`, thereby removing the limit on ALMs in the front-end.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchAgent.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchAgent.java
@@ -31,6 +31,9 @@ import java.lang.instrument.ClassFileTransformer;
 import java.lang.instrument.Instrumentation;
 import java.lang.instrument.UnmodifiableClassException;
 import java.security.ProtectionDomain;
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.Optional;
 
 public final class CommunityBranchAgent {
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchAgent.java
```diff
@@ -88,4 +91,37 @@ private static void redefineEdition(Instrumentation instrumentation) throws Clas
...
+        public String getMethodBody() {
+            return methodBody;
+        }
+    }
+
+    private enum Edition {
+        CE, WEB;
+
+        private static Optional<Edition> fromString(String input) {
+            return Arrays.stream(values()).filter(v -> v.name().toLowerCase(Locale.ENGLISH).equals(input)).findFirst();
```
YTN01, line 123: Will `input` ever come non-lower-cased? 
Does it make sense to first lowercase it before performing lookup?
```suggestion
      return Optional.ofNullable(input)
                     .map(i -> i.toLowerCase(Locale.ENGLISH))
                     .flatMap(i -> Arrays.stream(values())
                                         .filter(v -> v.name()
                                                       .toLowerCase(Locale.ENGLISH)
                                                       .equals(i))
                                         .findFirst());
```

src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchAgent.java
```diff
@@ -88,4 +91,37 @@ private static void redefineEdition(Instrumentation instrumentation) throws Clas
...
+        public String getMethodName() {
+            return methodName;
+        }
+
+        public String getMethodBody() {
+            return methodBody;
+        }
+    }
+
+    private enum Edition {
```
YTN01, line 119: I see this is a private inner class but, any reason to not make it static?

YTN01 (approved):
minor comments.

src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchAgent.java
mc1arke: The installation instructions only have the values specified in lower case and startup will fail-fast if the values don't match so I'm not too concerned about handling mixed-case. If this does become an issue longer-term then this can be simplified to the following to match case insentiviely:
```java
        return Arrays.stream(values).filter(v -> v.name().equalsIgnoreCase(input)).findFirst();
```

src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchAgent.java
mc1arke: Enums are always implicitly static so there's no need to explicitly declare them as static.

Issue #377, opened 2021-07-01
GitLab: mono repo support


qqshfox:
Hi,

Thank you for the awesome work. I was trying to use pull request decoration in our monorepo which is hosted on a self-hosted GitLab. However it didn't seem to support this usage. The decorator will delete all notes despite which project they belong to. I did find a merged PR to support GitHub but no GitLab yet. I'm wondering if there's a plan to support pull request decoration for gitlab monorepo?

I'm reading the code and it seems to lack project key for the notes created by decorator. Thinking to add the project key as a part of analysis result and file comments, and extract them by parsing note text for the next run. But not really a fan of parsing raw text. Not sure if there's other way to do this. Would you give me some suggestions? Thanks.

mc1arke:
Hi, thanks for the question!

As I'd done some initial work on the Gitlab decorator to retain and resolve discussions rather than deleting and recreating comments, which parsed the issue ID from the link we put in each comment, I've tidied up and pushed that work in #382. This has introduced the pattern you're asking about of parsing text, and allowed it to be performed fairly trivially whilst retrieving issue IDs, so I've pushed a change for it in #383. Please take a look and provide any comments, and if possible try out a the Snapshot build for that second MR and provide any feedback on how it works.

qqshfox:
Thanks a lot for implementing this so fast! I've tested #383 and it works great 👍 .

1. Creating summaries ✅ 
2. Creating hotspots ✅ 
3. Resolving summaries without comments ✅ 
4. Resolving hotspots without comments ✅ 
5. Leaving summaries with comments open ✅ 
6. Leaving hotspots with comments open ✅ 
7. No redundant SonarQube comments ✅ 

One minor suggestion only, would you add a project name in the summaries please, so that we could tell which sonarqube project it belongs to easily instead of hovering the `View in SonarQube` link to look for the project key? Thanks.

mc1arke:
Thanks for the comprehensive testing! As you've suggested, I've added the project ID into all comments (summary and individual issues), and issue IDs into the issue comments.

qqshfox:
Thank you so much! It looks perfect!

mc1arke:
Release in v1.9.0 of the plugin

mc1arke:
Backported to 1.8.1 of the plugin for Sonarqube 8.9 support.

PR #383, opened 2021-07-07
#377 Allow decorating Gitlab Mono Repository Merge Requests

mc1arke:
The Gitlab decorator treated any comment posted by the current Gitlab user as having been submitted for the current Sonarqube project, and therefore resolved any threads that no active issue could be found for in the current analysis. This prevented the decorator being used for a mono-repo as issues from decorations for other projects based in the repo would be resolved as subsequent projects completed their analysis.

The decorator has been changed to include the project ID in the details extracted from the existing comments, and to filter out any comments with project keys that do not match the project key the analysis is being performed for, before closing any remaining discussions, so that comments from other Sonarqube projects are not prematurely closed.


mc1arke:
Note: this is based on the commit in #382 so should only be merged after review and merge of #382

PR #404, opened 2021-08-03
it was misleading in readme

regentov:
Option name was misleading and doesn't work if typed as in instruction.


mc1arke:
Merged, thanks for the contribution!

PR #435, opened 2021-09-03
Hide outdated summary comments on GitHub

mc1arke:
When multiple Sonarqube analyses are performed against a Github Pull Request, each analysis generates a summary comment but none of the previous comments are cleaned-up on any subsequent analysis. To prevent a Pull Request becoming cluttered with summary comments, a search is being performed for all comments made by the current user, and all those comments are being minimized to prevent them being shown in-full on the Github user interface. As code level annotations are not treated as comments, this change has no impact on the individual findings within a Pull Request, only on the post-scan summary comment.


mc1arke:
Proposing this supercedes #323

sumeetp26:
Hi @mc1arke , the 1.8.2 branch plugin with Sonarqube 8.9.8 I see the comment gets minimised. Is there a way we can update the existing comment or delete the previous comment?

Issue #290, opened 2020-12-22
PR decoration not working against on-prem TFS 2018 update 3 - due to incorrect api-version value


arcdev:
**Describe the bug**
PR decoration fails against TFS 2018 update 3 because the default value for the REST service `api-version` is `6.0-preview.1`.
It appears to be expecting the **Prepare Analysis** Sonar Scanner task to set the `sonar.pullrequest.vsts.apiVersion` property accordingly.  However, this value is not set.

**To Reproduce**
1. Create a build in TFS & configure as a Gated Build
2. Add **Prepare Analysis Configuration** task; configure normally.
3. create a PR (with something for SonarQube to comment on)
4. (Gated Build runs with analysis)
5. SonarQube logs indicate decorations, but fails to publish them to the PR.

**Expected behavior**
SonarQube decorations appear in the PR.

**Software Versions**
 - SonarQube Version: Community Edition - Version 8.5.1 (build 38104)
 - Plugin Version: 1.6.0
 - SonarQube extension for TFS (sonar-scanner-vsts) 4.15.0 [marketplace link](https://marketplace.visualstudio.com/items?itemName=SonarSource.sonarqube)

**Workaround**
Manually set the property in the **Prepare Analysis** tasks, e.g.
`sonar.pullrequest.vsts.apiVersion=4.1-preview.1`

**Additional notes**
log snippet attached with some values updated for public consumption

[sonarQube-issue-log-01.txt](https://github.com/mc1arke/sonarqube-community-branch-plugin/files/5731036/sonarQube-issue-log-01.txt)

mc1arke:
@arcdev this sounds like it's working as expected: you're supposed to pass the additional parameter if you're running a versions of VSTS that requires a different API version

PR #438, opened 2021-09-07
Azure devops api version

jbaehr-phoenis:
Adds an option to configure the api version of azure devops.
This was prviously not possible and would lead to problems/errors.
This should also fix the problem in issue #290 .

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
@@ -134,7 +134,7 @@ public class CommunityBranchPlugin implements Plugin, CoreExtension {
                     PropertyDefinition.builder(AZURE_DEVOPS_API_VERSION)
                             .category(CoreProperties.CATEGORY_GENERAL)
                             .subCategory(CoreProperties.SUBCATEGORY_GENERAL)
-                            .defaultValue("6.0")
+                            .defaultValue("4.1")
                             .onQualifiers(Qualifiers.APP)
                             .name("Azure DevOps API Version")
                             .description("API Version to submit with any Azure DevOps API request")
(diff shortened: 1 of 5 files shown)


mc1arke:
Thanks for your suggested contribution. As the Azure DevOps decorator has been significantly overhauled since the issue that you've linked to was raised, can you confirm the scenario where you’re still encountering problems please? I'm generally against adding user-interface options for configuring the Pull Request decorators since they don't then sit alongside the main decorator configuration options so cause a poor user experience, so would like to understand the issue being resolved here.

dhaas-phoenis:
Our Azure DevOps Server 2020 (on-premise) does not accept api requests with api-version=4.1. That is why we added this option. I understand why you are against adding this user-interface option. What would be a better solution?

See attached file:
[sonarqube_ce.log](https://github.com/mc1arke/sonarqube-community-branch-plugin/files/7126065/sonarqube_ce.log)

mc1arke:
It's odd that 2020 is giving the response you've provided, as the v4 API has existed since 2018 - https://docs.microsoft.com/en-us/azure/devops/integrate/concepts/rest-api-versioning?view=azure-devops#supported-versions. Can you confirm if your server is generally up-to-date with latest patches?

I wonder whether the use of 4.1 rather than 4.0 is causing any issues, and also whether pinning to 3.1 would provide a wider range of support across all versions of Azure Devops currently available.

dhaas-phoenis:
Yes, I can confirm. We use the latest available patches. It's Azure DevOps Server 2020 Update 1.1.

Not every API needs the "-preview"-flag. But the Pull Request Statuses API does. No matter which version I use. Tested from 3.0 to 6.1. Every version needs it. But Azure DevOps Services (cloud) only needs the preview flag for api-version 6.1. Seems to be something on-premise specific.

Should we cancel this PR and create an issue instead?! We can still use our own forked version that works for us until this is fixed.

mc1arke:
Thanks for the confirmation. I'd agree that the fix for this is slightly different from your proposed change since it requires using a preview version for a few selected endpoint calls rather than allowing configuring a version for all endpoint calls. Could you therefore raise an issue - or an alternative PR - to cover this, and close this PR?

PR #441, opened 2021-09-13
Apply appropriate casing to CommentThreadStatus property names

### LINKED ISSUE #439: Azure DevOps : exception deserializing comment thread
Caused by: com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `com.github.mc1arke.sonarqube.plugin.almclient.azuredevops.model.enums.CommentThreadStatus` from String "wontFix": not one of the values accepted for Enum class: [fixed, bydesign, active, wontfix, unknown, closed, pending]

jcuzzi:
Fixes #439

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/almclient/azuredevops/model/enums/CommentThreadStatus.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/almclient/azuredevops/model/enums/CommentThreadStatus.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
  * The status of a comment thread
+ * Enum names match those here: https://docs.microsoft.com/en-us/rest/api/azure/devops/git/pull-request-threads/get?#commentthreadstatus
  */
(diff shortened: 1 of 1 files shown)


mc1arke:
Is the correct fix for this not to update the `@JsonProperty` values in https://github.com/mc1arke/sonarqube-community-branch-plugin/blob/master/src/main/java/com/github/mc1arke/sonarqube/plugin/almclient/azuredevops/model/enums/CommentThreadStatus.java to match the values in https://docs.microsoft.com/en-us/javascript/api/azure-devops-extension-api/commentthreadstatus?

jcuzzi:
Either way works.  I think this way makes it more fool proof, but it is up to you.

mc1arke:
Part of the reason I'm suggesting changing the enum is that any serialised value produced by the service wont work after your change, since they wont match the possibly values supported by the API. We don't currently serialise these enum values so it's not currently causing an issue, but it's something we should fix in the right way if we can.

mc1arke:
Merged, thanks for the quick turn-around on the fix.

Issue #445, opened 2021-09-16
[Gitlab] Error while decorating the request 


sfaqer:
**Describe the bug**
Found error in CE logs, MR not decorated

**Expected behavior**
MR successful decorated

**Software Versions**
 - SonarQube Version: 9.0.1 (build 46107)
 - Plugin Version: 1.9.0
 - Docker image: mc1arke/sonarqube-with-community-branch-plugin:9.0-community

**Additional context**
```
sonarqube_1  | 2021.09.15 08:05:11 INFO  ce[AXvofOU_UQuhcpSys54E][c.g.m.s.p.c.p.PullRequestPostAnalysisTask] using pull request decorator com.github.mc1arke.sonarqube.plugin.ce.pullrequest.gitlab.GitlabMergeRequestDecorator
sonarqube_1  | 2021.09.15 08:05:12 ERROR ce[AXvofOU_UQuhcpSys54E][c.g.m.s.p.a.g.GitlabRestClient] Gitlab response status did not match expected value. Expected: 201\nHttpResponseProxy{HTTP/1.1 400 Bad Request [Server: nginx, Date: Wed, 15 Sep 2021 08:05:12 GMT, Content-Type: application/json, Content-Length: 107, Connection: keep-alive, Cache-Control: no-cache, Vary: Origin, X-Content-Type-Options: nosniff, X-Frame-Options: SAMEORIGIN, X-Request-Id: 01FFM7T0RH6WGZS0KPXTMBQ44E, X-Runtime: 0.074482] ResponseEntityProxy{[Content-Type: application/json,Content-Length: 107,Chunked: false]}}\n{"message":"400 Bad request - Note {:line_code=\u003e[\"can't be blank\", \"must be a valid line code\"]}"}
sonarqube_1  | 2021.09.15 08:05:12 ERROR ce[AXvofOU_UQuhcpSys54E][o.s.c.t.p.a.p.PostProjectAnalysisTasksExecutor] Execution of task class com.github.mc1arke.sonarqube.plugin.ce.pullrequest.PullRequestPostAnalysisTask failed
sonarqube_1  | java.lang.IllegalStateException: An unexpected response code was returned from the Gitlab API - Expected: 201, Got: 400
sonarqube_1  |  at com.github.mc1arke.sonarqube.plugin.almclient.gitlab.GitlabRestClient.validateResponse(GitlabRestClient.java:230)
[...]

PR #446, opened 2021-09-16
Added encoding for entity sending to gitlab.

sfaqer:
fixed: #445 

Currently the Gitlab decorator does not set an encoding when sending the merge request decoration details to Gitlab, which results in some characters - such as the Cyrillic alphabet - being rendered incorrectly on Gitlab, or if that character found on old \ new path position occurs error in API. Explicitly specifying UTF-8 as the encoding overcome this.

p.s restored fix from: https://github.com/mc1arke/sonarqube-community-branch-plugin/commit/4a9918dfa34376346ae98fc7ada4a7b4babfa0eb


mc1arke:
Is there a test we can add to prevent this regressing again during any future refactor?

mc1arke:
Superseded by #472  which covers more scenarios and contains tests to safeguard the functionality. Thanks for your suggested contribution.

Issue #454, opened 2021-09-27
Plugin does not work with Sonarqube 9.1


axylophon:
Hi we updated our SonarQube and the branch Plugin. So far so good, but the Devops PR decoration is not working anymore. An exception is thrown during the PR decoration. The same error is also thrown when accessing the DevOps Platform Integration menu in the SonarQube UI. Is this caused by the SonarQube 9.1 Version?

```
java.lang.NoSuchMethodError: org.sonar.db.alm.setting.AlmSettingDto.getPersonalAccessToken()Ljava/lang/String;
	at com.github.mc1arke.sonarqube.plugin.almclient.azuredevops.DefaultAzureDevopsClientFactory.createClient(DefaultAzureDevopsClientFactory.java:53)
	at com.github.mc1arke.sonarqube.plugin.ce.pullrequest.azuredevops.AzureDevOpsPullRequestDecorator.createClient(AzureDevOpsPullRequestDecorator.java:79)
	at com.github.mc1arke.sonarqube.plugin.ce.pullrequest.azuredevops.AzureDevOpsPullRequestDecorator.createClient(AzureDevOpsPullRequestDecorator.java:60)
	at com.github.mc1arke.sonarqube.plugin.ce.pullrequest.DiscussionAwarePullRequestDecorator.decorateQualityGateStatus(DiscussionAwarePullRequestDecorator.java:67)
	at com.github.mc1arke.sonarqube.plugin.ce.pullrequest.PullRequestPostAnalysisTask.finished(PullRequestPostAnalysisTask.java:160)
	at org.sonar.ce.task.projectanalysis.api.posttask.PostProjectAnalysisTasksExecutor.executeTask(PostProjectAnalysisTasksExecutor.java:110)
	at org.sonar.ce.task.projectanalysis.api.posttask.PostProjectAnalysisTasksExecutor.finished(PostProjectAnalysisTasksExecutor.java:101)
	at org.sonar.ce.task.step.ComputationStepExecutor.executeListener(ComputationStepExecutor.java:91)
	at org.sonar.ce.task.step.ComputationStepExecutor.execute(ComputationStepExecutor.java:63)
	at org.sonar.ce.task.projectanalysis.taskprocessor.ReportTaskProcessor.process(ReportTaskProcessor.java:81)
	at org.sonar.ce.taskprocessor.CeWorkerImpl$ExecuteTask.executeTask(CeWorkerImpl.java:212)
[...]

PR #455, opened 2021-09-27
#454: Add support for Sonarqube 9.1

mc1arke:
Sonarqube 9.1 encrypts the sensitive fields for DevOps decorators, such as private keys and personal access tokens. To allow the decorators to access these properties, the relevant methods now require encryption details to be provided, so the encryption handlers are now being loaded from the internal Sonarqube settings and passed to the appropriate target methods.

A change is also required in `SetBindingsAction` to pass additional fields for recording on the audit log when settings are saved. No additional settings are being added for the auditing as community edition does not provide a user interface for accessing audit logs, and uses a default cleaner task for purging logs after 30 days.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ repositories {
     }
 }
 
-def sonarqubeVersion = '9.0.0.45539'
+def sonarqubeVersion = '9.1.0.47736'
 def sonarqubeLibDir = "${projectDir}/sonarqube-lib"
 def sonarLibraries = "${sonarqubeLibDir}/sonarqube-${sonarqubeVersion}/lib"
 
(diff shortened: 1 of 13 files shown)


KrylixZA:
Hey @mc1arke 

I've updated my SonarQube 9.1 on-prem instance to use the 1.10.0 snapshot plugin. All seems well with Azure DevOps :)

I did notice that the comment that is being added as a summary to the PRs with the images renders the images like this:
image

This is because in my case the SonarQube server is not accessible through Azure DevOps directly (have to go through on-prem self-hosted agents to get there). While it's not an issue as the text is clearly readable, a fallback image or maybe no image would be cool. 

Anyway, really appreciate you making this change!

KTannenberg:
Hi @mc1arke, can also confirm that 1.10.0-SNAPSHOT version is working great (GitHub Enterprise + Sonarqube 9.1.0.47736):
image

Is there rough ETA on when new version will be released?

stevehipwell:
@mc1arke any news on an ETA for v1.10.0? SonarQube [v9.2.0](https://github.com/SonarSource/sonarqube/releases/tag/9.2.0.49834) has just dropped.

PR #487, opened 2021-11-17
Add configuration option to disable issue comments on PR decoration

teake:
For PRs with lots of issues, it can be quite noisy to post comments each raised issue. This change adds the option to disable comments for issues, leaving only the summary comment. The option can be found in both the global settings ("General -> General" for lack of a better place), and in the project settings.

Note that the project settings currently do not have an effect due to [SONAR-13711](https://jira.sonarsource.com/browse/SONAR-13711).

Also, this will only affect Azure and GitLab integrations.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
@@ -25,6 +25,7 @@ import com.github.mc1arke.sonarqube.plugin.almclient.github.DefaultGithubClientF
 import com.github.mc1arke.sonarqube.plugin.almclient.github.v3.RestApplicationAuthenticationProvider;
 import com.github.mc1arke.sonarqube.plugin.almclient.gitlab.DefaultGitlabClientFactory;
 import com.github.mc1arke.sonarqube.plugin.ce.CommunityReportAnalysisComponentProvider;
+import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.DiscussionAwarePullRequestDecorator;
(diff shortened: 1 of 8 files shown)


mc1arke:
A few initial comments:
1. I don't really want to have non-standard properties provided by this plugin as I generally want users to follow Sonarsource official documentation
2. I definitely don't want properties that are only effective for some implementations or are unclear about whether they're applied
3. I'm not sure I agree with the use case: you could achieve the same thing by not enabling decoration and having your build tool report its pipeline result.

If the problem is that you have a lot of issues being reported then you could look to implement a limit in the decoration to only apply annotations for a capped number of findings plus an additional entry on the summary comment to make it clear that not all issues have been annotated.

teake:
> A few initial comments:
> 
>     1. I don't really want to have non-standard properties provided by this plugin as I generally want users to follow Sonarsource official documentation
> 
>     2. I definitely don't want properties that are only effective for some implementations or are unclear about whether they're applied

Fair enough. Though I would have appreciated knowing this upfront (the second point is kinda obvious, the first one isn't) -- perhaps worth putting this in a "Contributing" section in README.md?

>     3. I'm not sure I agree with the use case: you could achieve the same thing by not enabling decoration and having your build tool report its pipeline result.

That's not quite the same thing IMHO. I would like to retain the summary comment, as it already provides a lot of information without having to visit SonarQube.

> If the problem is that you have a lot of issues being reported then you could look to implement a limit in the decoration to only apply annotations for a capped number of findings plus an additional entry on the summary comment to make it clear that not all issues have been annotated.

Would that not require a new config setting for the limit? That would be in conflict with your first point.

But in any case, thanks for the feedback. Since I cannot find the time to bring this PR up to spec, feel free to close it.

Issue #523, opened 2021-12-24
Github Actions environment variables are not auto-detected by the scanner


pcolmer:
**Describe the bug**
1. Performing a straightforward analysis on a branch commit does not store the analysis under that branch's name. Instead, it gets saved under the name "develop".
2. Performing a pull request does not store the analysis under the "Pull Request" tab on Sonar. Instead, it gets saved against the name "develop".

**To Reproduce**
1. Create an empty database on PostgreSQL. I'm using AWS RDS for this purpose.
2. Start up a Docker container with the image that embeds the plugin. For the purpose of gaining a public IP address, I'm using the facility within Docker Desktop to spin up the container within ECS Fargate but I have no reason to believe that this is changing the behaviour of the container.
3. Perform a GitHub Actions-triggered scan on a branch commit
4. Perform a GitHub Actions-triggered scan on a pull request

**Expected behavior**
1. I would expect the "+" symbol next to the branch name in the Sonar UI to give me a list of branches and not a pop-up explaining how to analyse branches.
2. I would expect commit-triggered analyses to be saved under the correct branch name.
3. I would expect PR-triggered analyses to be saved under the correct branch name and handled as a pull request.

**Software Versions**
 - SonarQube Version: Version 9.2.3 (build 50713)
 - Plugin Version: 1.10.0 (mc1arke/sonarqube-with-community-branch-plugin:9.2-community)

**Additional context**
docker-compose file used to run the Docker container:
```
version: "3"

# Specify the VPC ID within the AWS account
x-aws-vpc: "vpc-12345678"

# Define the Sonarqube service.
services:
  sonarqube:
    image: mc1arke/sonarqube-with-community-branch-plugin:9.2-community
    deploy:
      replicas: 1
      resources:
        limits:
          cpus: '0.5'
          memory: 4096M
    ulimits:
      nofile:
        soft: 65536
        hard: 65536
    environment:
      - sonar.jdbc.url=jdbc:postgresql://mydatabase.us-east-1.rds.amazonaws.com:5432/sonar
      - sonar.jdbc.username=sonar
      - sonar.jdbc.password=sonarpassword
      - sonar.search.javaAdditionalOpts=-Dnode.store.allow_mmap=false
    volumes:
      - sonarqube_data:/opt/sonarqube/data
      - sonarqube_logs:/opt/sonarqube/logs
      - sonarqube_conf:/opt/sonarqube/conf
    ports:
      - target: 9000
[...]

PR #526, opened 2021-12-29
#523: Add auto-detection for all CIs in Sonarqube commercial editions

mc1arke:
The plugin previously only provided support for auto-detecting and configuring the scanner properties for a Pull Request in Azure Devops and a Merge Request or Branch in Gitlab CI. The Sonarqube documentation also stated that Bitbucket Pipelines, Github Actions, CodeMagic, Jenkins Branch API, and Cirrus CI could also be used to auto-discover Pull Request or Branch information although the plugin did not provide these.

This change adds support for detecting these additional CIs based on the various environment variables they provide, and to auto-configure Pull Request or Branch parameters in the scanner when a suitable build job is detected.

Includes the general clean-up of the creation of Branch and Pull Request configuration to force fail-fast behaviour where target branches are not provided or can't be matched against known branches, to ensure the correct reference branch is selected for Pull Request analysis, and to force an error to be displayed if a user mixes Pull Rrequest and Branch parameters in their launch properties.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/scanner/autoconfiguration/GithubActionsAutoConfigurer.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/scanner/autoconfiguration/GithubActionsAutoConfigurer.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2022 Michael Clarke
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ */
+package com.github.mc1arke.sonarqube.plugin.scanner.autoconfiguration;
+
+import com.github.mc1arke.sonarqube.plugin.scanner.BranchAutoConfigurer;
+import com.github.mc1arke.sonarqube.plugin.scanner.BranchConfigurationFactory;
+import org.apache.commons.lang.StringUtils;
+import org.sonar.api.utils.System2;
+import org.sonar.scanner.scan.branch.BranchConfiguration;
+import org.sonar.scanner.scan.branch.ProjectBranches;
+
+import java.util.Optional;
+
+public class GithubActionsAutoConfigurer implements BranchAutoConfigurer {
+
+    private static final String PULL_REQUEST_REF_PREFIX = "refs/pull/";
+    private static final String PULL_REQUEST_REF_POSTFIX = "/merge";
+
+    private final BranchConfigurationFactory branchConfigurationFactory;
+
+    public GithubActionsAutoConfigurer(BranchConfigurationFactory branchConfigurationFactory) {
+        this.branchConfigurationFactory = branchConfigurationFactory;
+    }
+
+    @Override
+    public Optional<BranchConfiguration> detectConfiguration(System2 system, ProjectBranches projectBranches) {
+        if (!Boolean.parseBoolean(system.envVariable("GITHUB_ACTIONS"))) {
+            return Optional.empty();
+        }
+
+        String ref = system.envVariable("GITHUB_REF");
+        if (StringUtils.isEmpty(ref)) {
+            return Optional.empty();
+        }
+
+        String sourceBranch = system.envVariable("GITHUB_HEAD_REF");
+        if (ref.startsWith(PULL_REQUEST_REF_PREFIX) && ref.endsWith(PULL_REQUEST_REF_POSTFIX) && StringUtils.isNotEmpty(sourceBranch)) {
+            String key = ref.substring(PULL_REQUEST_REF_PREFIX.length(), ref.length() - PULL_REQUEST_REF_POSTFIX.length());
+            String targetBranch = system.envVariable("GITHUB_BASE_REF");
+            return Optional.of(branchConfigurationFactory.createPullRequestConfiguration(key, sourceBranch, targetBranch, projectBranches));
+        } else {
+            String branch = system.envVariable("GITHUB_REF_NAME");
+            return Optional.of(branchConfigurationFactory.createBranchConfiguration(branch, projectBranches));
+        }
+    }
+}
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020-2021 Michael Clarke
+ * Copyright (C) 2020-2022 Michael Clarke
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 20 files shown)


pcolmer:
I've incorporated the plugin and run a pull request against the develop branch of a project. I'm seeing two potential areas of concern within the SQ server:

1. The scan is recorded as a branch and not as a pull request.
2. The branch name is evelop.

I can see the following in the output from the scan action:

```Auto detected BRANCH configuration with source evelop using com.github.mc1arke.sonarqube.plugin.scanner.autoconfiguration.GithubActionsAutoConfigurer```

and, later on:

```INFO: Branch name: evelop```

**Comment by pcolmer** [comment]
@mc1arke if there is any further testing I can do to help, please just let me know. Thanks.

**Comment by mc1arke** [comment]
@pcolmer Apologies for the extremely delayed response. My test for the Github refs had an incorrect format for the value, which then resulted in my implementation attempting to drop too many characters. I've corrected the test and the implementation. Could you retest this please?

**Comment by pcolmer** [comment]
@mc1arke Thanks for the update.

The code is now correctly extracting the branch name `develop`. However, the pull request is still being detected as a BRANCH:

> INFO: Auto detected BRANCH configuration with source develop using com.github.mc1arke.sonarqube.plugin.scanner.autoconfiguration.GithubActionsAutoConfigurer

Looking at the values for the various GITHUB environment variables while the pull request is being processed, it looks like there may be an error in the GithubActionsAutoConfigurer code.

When I'm running the pull request action, these are the key values:
```
GITHUB_REF=refs/heads/develop
GITHUB_HEAD_REF=pjc-test
```

Since `GITHUB_REF` does not start with `refs/pull/`, the code automatically defers to it being a branch not a pull request. Also, it doesn't end with `/merge`.

I hope that helps.

**Comment by mc1arke** [comment]
@pcolmer Can you confirm this is a run triggered from the Pull Request, not a run for the push to the source branch? As you've pointed out, the `GITHUB_REF` doesn't match the Pull Request format the code is looking for, which doesn't seem to match my interpretation of the documentation of the variable in https://docs.github.com/en/actions/learn-github-actions/environment-variables

File: src/main/java/com/github/mc1arke/sonarqube/plugin/scanner/autoconfiguration/GithubActionsAutoConfigurer.java [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+            return Optional.empty();
+        }
+
+        String sourceBranch = system.envVariable("GITHUB_HEAD_REF");
+        if (branch.startsWith(PULL_REQUEST_REF_PREFIX) && branch.endsWith(PULL_REQUEST_REF_POSTFIX) && StringUtils.isNotEmpty(sourceBranch)) {
+            String key = branch.substring(PULL_REQUEST_REF_PREFIX.length(), branch.length() - PULL_REQUEST_REF_POSTFIX.length());
+            String targetBranch = system.envVariable("GITHUB_BASE_REF");
+            return Optional.of(branchConfigurationFactory.createPullRequestConfiguration(key, sourceBranch, targetBranch, projectBranches));
+        } else {
+            return Optional.of(branchConfigurationFactory.createBranchConfiguration(branch.substring(11), projectBranches));
```
usmonster, line 58: You can use `GITHUB_REF_NAME`:
```suggestion
        String branch = system.envVariable("GITHUB_REF_NAME");
        if (StringUtils.isEmpty(branch)) {
            return Optional.empty();
        }

        String sourceBranch = system.envVariable("GITHUB_HEAD_REF");
        if (branch.startsWith(PULL_REQUEST_REF_PREFIX) && branch.endsWith(PULL_REQUEST_REF_POSTFIX) && StringUtils.isNotEmpty(sourceBranch)) {
            String key = branch.substring(PULL_REQUEST_REF_PREFIX.length(), branch.length() - PULL_REQUEST_REF_POSTFIX.length());
            String targetBranch = system.envVariable("GITHUB_BASE_REF");
            return Optional.of(branchConfigurationFactory.createPullRequestConfiguration(key, sourceBranch, targetBranch, projectBranches));
        } else {
            return Optional.of(branchConfigurationFactory.createBranchConfiguration(branch, projectBranches));
```

mc1arke: I can't use `GITHUB_REF_NAME` at the point you've suggested as line in 53 depends on knowing the full ref value. It would be possible to use the `GITHUB_REF_NAME` environment variable in line 54 and 58, although given we know the format of the ref at that point I'm not sure if this provides a huge amount of value.

src/main/java/com/github/mc1arke/sonarqube/plugin/scanner/autoconfiguration/GithubActionsAutoConfigurer.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
```
usmonster, line 60: **Edit** You can use `GITHUB_REF_NAME`:
```suggestion
        String ref = system.envVariable("GITHUB_REF");
        if (StringUtils.isEmpty(ref)) {
            return Optional.empty();
        }
        
        String sourceBranch = system.envVariable("GITHUB_HEAD_REF");
        if (StringUtils.isNotEmpty(sourceBranch)) {
            String key = ref.substring(PULL_REQUEST_REF_PREFIX.length(), ref.length() - PULL_REQUEST_REF_POSTFIX.length());
            String targetBranch = system.envVariable("GITHUB_BASE_REF");
            return Optional.of(branchConfigurationFactory.createPullRequestConfiguration(key, sourceBranch, targetBranch, projectBranches));
        }
        
        String branch = system.envVariable("GITHUB_REF_NAME");
        return Optional.of(branchConfigurationFactory.createBranchConfiguration(branch, projectBranches))
```

src/main/java/com/github/mc1arke/sonarqube/plugin/scanner/autoconfiguration/GithubActionsAutoConfigurer.java
usmonster: Sorry, I selected incomplete context and then commented too soon! I re-commented with that I had in mind, but it at least avoids the potential bug in `branch.substring(11)` which would lead to something unexpected in the case where the ref is a tag.

pcolmer:
@mc1arke It is definitely a pull request. The trigger action is `pull_request_target`.

It may be worth considering a different logic to tell the difference between a pull request and a branch operation. According to https://docs.github.com/en/actions/learn-github-actions/environment-variables, `GITHUB_HEAD_REF` is only set when the event that triggers a workflow run is either `pull_request` or `pull_request_target`. Therefore, if you can retrieve a value for `GITHUB_HEAD_REF`, it surely has to be a pull request and if you can't retrieve a value, it must be a branch operation?

pcolmer:
@mc1arke I take your point about the documented examples for `GITHUB_REF`. I'll raise a documentation issue with the GitHub team as that is clearly not what is going on at the moment.

mc1arke:
> It may be worth considering a different logic to tell the difference between a pull request and a branch operation. According to https://docs.github.com/en/actions/learn-github-actions/environment-variables, `GITHUB_HEAD_REF` is only set when the event that triggers a workflow run is either `pull_request` or `pull_request_target`. Therefore, if you can retrieve a value for `GITHUB_HEAD_REF`, it surely has to be a pull request and if you can't retrieve a value, it must be a branch operation?

This wouldn't allow the plugin to gather enough information: to perform Pull Request decoration the Pull Request ID is needed. None of the variables you can in your environment dump give an indication of the Pull Request ID, and nothing in the documentation indicates another variable that can be used to retrieve or calculate this value.

I'm not sure that `pull_request_target` is a valid event to be watching for based on https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target. It states

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.

which reads to me as if you're acting on the target branch of the pull request, not the actual pull request.

pcolmer:
@mc1arke I'll change the action to be `pull_request` and see if that changes the behaviour.

pcolmer:
@mc1arke Changing the trigger action to `pull_request` does change the variable values that are passed and the plugin correctly identifies it as a pull request. Thank you for your support and patience :)

PR #541, opened 2022-02-01
Fix #77 approve bitbucket PRs when configured.

### LINKED ISSUE #77: Approve PR that passed quality gate
This feature will allow to configure the PR to make it required to pass the quality gate. It is only necessary to add a comment user to the required reviewers.

nfalco79:
This feature requires a write access configured in OAuth consumers.
Add a new configuration parameter under ALMSettings/Bitbucket to enable or not this behavior (default false). When enabled and sonar analysis a pull request branch the logged user will approve/disapprove current PR based on quality gate result.

Before spend time to implement test case please let me know if you are yet interested on this (#77) feature

immagine
In picture seems I can not change how "bitbucket" sub category appears, adding a description for example. It looks like something that depends on sonarqube server.


mc1arke:
You can't add options on the MR/PR configuration screens given the way they're built. I'm also not a fan of adding configuration on other screens as it makes it difficult for users to find all the configuration options they're supposed to set. Is approvals something that Sonarqube offers in its commercial offerings?

nfalco79:
Do you prefer a dedicated page? But this page shold refer only to bitbucket action

Issue #298, opened 2021-01-29
NPE Analysis failure against core branch when PR analysis marked as CANCELLED


jgrant216:
TL;DR: A PR against `develop` had its analysis task marked as `CANCELLED`.  There was only one analysis, none prior.  `develop` analysis fails due to missing `analysisDate` element in child `CANCELLED` PR.

Workflow...

1. PR-71 submitted against core branch `develop`.
2. PR-71 analysis background task on SQ marked as `CANCELLED`.
3. CI for PR-71 does not validate SQ was successful, PR-71 was merged to `develop`.
4. `develop` CI runs, `EXECUTION FAILURE` because call to http://SQHost/api/project_pull_requests/list?project=thisprojectkey returned analysis for key `71` (PR-71) which is missing `analysisDate` element.

CANCELLED PR Analysis Sample JSON response...
      {
         "key":"71",
         "title":"feature/testingDIT",
         "branch":"feature/testingDIT",
         "base":"develop",
         "status":{
            "bugs":0,
            "vulnerabilities":0,
            "codeSmells":0
         },
         "target":"develop"
      },

Good PR Analysis Sample JSON response...
      {
         "key":"65",
         "title":"feature/db_changes_with_vault_only",
         "branch":"feature/db_changes_with_vault_only",
         "base":"develop",
         "status":{
            "qualityGateStatus":"OK",
            "bugs":0,
            "vulnerabilities":0,
            "codeSmells":0
         },
         "analysisDate":"2021-01-22T13:00:19-0500",
         "target":"develop"
      },

Environment:

- Community EditionVersion 8.4.2 (build 36762)
- Community Branch Plugin 1.5.0
- CLI 4.6.0.2311

Stack trace with `-X` including reported plugin information...
13:33:55.831 INFO: Load/download plugins (done) | time=9431ms
13:33:56.417 DEBUG: Plugins:
13:33:56.417 DEBUG:   * Python Code Quality and Security 2.13.0.7236 (python)
13:33:56.417 DEBUG:   * SonarCSS 1.2.0.1325 (cssfamily)
13:33:56.417 DEBUG:   * JaCoCo 1.1.0.898 (jacoco)
13:33:56.417 DEBUG:   * SonarGo 1.6.0.719 (go)
13:33:56.417 DEBUG:   * SonarKotlin 1.5.0.315 (kotlin)
[...]

PR #548, opened 2022-02-16
#298: Prevent NullPointerException when branch with cancelled analysi…

gauee:
Apply the fix `#298: Prevent NullPointerException when branch with cancelled analysis is rescanned` to release/1_3


mc1arke:
I'm not planning another release of 1.3 given a newer LTS of Sonarqube is available with a version of the plugin that does not contain this bug.

PR #552, opened 2022-02-28
Allow multiple reports in one Bitbucket PR

MrMightyNighty:
Since the previous implementation had a static report key, Bitbucket wasn't able to show multiple reports on its "Integration" window. To solve this issue the project key is used as the report key since report keys have to be unique or otherwise the reports will be overwritten. This was tested on a 8.9 LTS on-prem instance and works perfectly fine.


MuhKuh7:
Any chance this feature can make it into the next release?

mc1arke:
Replaced by #578 to fix merge conflicts

Issue #569, opened 2022-04-04
PR Decoration not working on SonarQube 9.4


juanrodicio:
**Describe the bug**
PR Decorator is not working on the latest SonarQube version.
Below you can see ce.log from sonarqube server in TRACE mode.

**Log**
`
[...]

PR #570, opened 2022-04-04
#569: Add support for release 9.4 of Sonarqube

mc1arke:
The latest release of Sonarqube switches from using Pico Container for
dependency injection to using Spring. The method used for detecting the
constructor to be used for injecting in a multi-constructor class varies
between the 2 frameworks, with Pico only looking for public constructors
but Spring looking for any visible constructor, and then excluding any
constructors that do not have the `@Autowired` annotation on a muti-
constructor class. This change adds the relevant annotation onto the two
classes impacted by this change in frameworks, thereby allowing for the
Compute Engine to have the appropriate components generated for
completing decoration of Pull Requests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020-2021 Michael Clarke
+ * Copyright (C) 2020-2022 Michael Clarke
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


xahhy:
So this will fix the issue of not show PR decoration for version 9.4 right? Will it be a new version of the plugin?

mc1arke:
> So this will fix the issue of not show PR decoration for version 9.4 right?

Yes. You can test this out by using the snapshot build of the plugin from this Pull Request: See the arifact section on https://github.com/mc1arke/sonarqube-community-branch-plugin/actions/runs/2092692348

> Will it be a new version of the plugin?

Yes. This will be a new minor version of the plugin

mc1arke:
I'm proposing this is superceded by #572 and #574 since they fix the constructor injection issue without making the plugin require Spring to be present, and therefore allows the plugin to remain backwards compatible with 9.1 and above of Sonarqube.

honzajscz:
I have tested the above-mentioned snapshot and PRs are decorated by the comments. However, a status check is not updated and PRs cannot be closed in Azure DevOps 
image

I have also noticed that the See the PR link is missing after the update. This is before update
image

Issue #285, opened 2020-12-08
Disable code coverage report on Gitlabci


LaSylv:
**Describe the bug**

Disable the coverage pushed to gitlab, as it is sometimes not correct and gitlab already provides this feature.

**Expected behavior**

It wold be nice to be able to disable the coverage pushed by the plugin through a specific property.

**Additional context**

As we already use the coverage through gitlab (more complete, showing which lines are covered or not), the sonar gitlab is a pain to set and never seem to be correct. We would like to disable it entirely when it's pushed to gitlab.

If this is somzthing you would consider, I can work on a PR

lucasoares:
@Glosur, what you mean about disabling coverage pushed to gitlab?

Is the coverage provided by the external job created by the plugin?

![image](https://user-images.githubusercontent.com/10624972/108893148-82fa3000-75ef-11eb-92ce-75569be0573c.png)

If so, I agree with you because the final coverage displayed on the MR page will be incorrect since it will show the average coverage between my test and the coverage of the new code.

For example, if I have one MR with only 1 changed line that doesn't is covered by tests, the widget will report an incorrect coverage of `(test coverage + sonar coverage) / 2`:

![image](https://user-images.githubusercontent.com/10624972/108892883-2ac32e00-75ef-11eb-8c36-c7c6ee85c907.png)

Glosur:
Yes, we are having the problem you're saying, that is heavily linked with https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/225


We use another gitlabci job to get the covergae from. Gitlab then also receives the coverage from the sonar plugin (the coverage is not the correct one, see PR linked)

In the end, even if the PR is fixed, I have to maintain two tools to have the coverage in the gitlabci job matching the one in sonar

lucasoares:
Even with #225 merged we are having this problem as well.

I just ran an analysis with sonarqube 9.5 and with plugin version 1.12.0 and the reported coverage in the SonarQube external job is the new code coverage and not the total coverage.

![image](https://user-images.githubusercontent.com/10624972/184224425-40e446b9-2a94-4803-a91d-c96dbd709642.png)

Karql:
Hi @mc1arke 

[...]

PR #225, opened 2020-08-06
Send total coverage back to GitLab instead of new coverage

teake:
Reworked version of #157 (and #145). Third time's the charm, right? 🤞

The coverage send back via the GitLab API should be the total coverage
of the project, not the coverage of the new code in the PR. The GitLab
docs explicitly mention this: 
https://docs.gitlab.com/ee/api/commits.html#post-the-build-status-to-a-commit

The coverage reported by decorating the PR via a comment is not affected;
this still shows both the total and new coverage.

Note that there is an open GitLab issue for natively displaying new coverage
for just the diff of a PR (https://gitlab.com/gitlab-org/gitlab/-/issues/20895),
but this is not yet implemented.

In addition, reporting coverage back is optional. So if there is no coverage
information, we do not report it back.

PR #577, opened 2022-04-10
Move report generation out of AnalysisDetails

mc1arke:
The access to metrics from a Pull Request analysis is exposed through an
`AnalysisDetails` instance, which also provides the ability to extract a
formatted report. As a number of the metrics used in the summary report
need to be retrieved through various additional DAOs, and as the
resolution of URLs for links and images requiring access to core
Sonarqube configuration, `AnalysisDetails` holds references to a high
number of classes from Sonarqube's core. Some of those core Sonarqube
classes are also referenced directly in some decorators which don't make
use of the summary report but need equivalent metrics to those shown in
the summary which means some searching logic is duplicated across the
plugin.

This change pulls the report generation into a `ReportGenerator` class,
with the report being an interim set of collected metrics that each
decorator can extract required information, or generate a formatted
report from.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
@@ -31,19 +31,17 @@ import com.github.mc1arke.sonarqube.plugin.almclient.gitlab.model.User;
 import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.AnalysisDetails;
 import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.DiscussionAwarePullRequestDecorator;
 import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.PostAnalysisIssueVisitor;
-import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.markup.FormatterFactory;
 import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.markup.MarkdownFormatterFactory;
+import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.report.AnalysisIssueSummary;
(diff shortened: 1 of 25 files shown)


src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
```diff
@@ -118,25 +116,18 @@ protected List<String> getCommitIdsForPullRequest(GitlabClient gitlabClient, Mer
...
-                URLEncoder.encode(analysis.getAnalysisProjectKey(), StandardCharsets.UTF_8.name()),
-                URLEncoder.encode(analysis.getBranchName(), StandardCharsets.UTF_8.name()));
-
             PipelineStatus pipelineStatus = new PipelineStatus("SonarQube",
                     "SonarQube Status",
                     analysis.getQualityGateStatus() == QualityGate.Status.OK ? PipelineStatus.State.SUCCESS : PipelineStatus.State.FAILED,
-                    dashboardUrl,
-                    coverage,
+                    analysisSummary.getDashboardUrl(),
+                    analysisSummary.getNewCoverage(),
```
Karql, line 130: Hi, @mc1arke this change is regresion for https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/225

mc1arke: Hi @Karql! Rather than placing similar comments at different points in the project, could you raise an issue stating the problem and replication steps, and provide any links to changes you believe have caused the issue at that point? This way other people have visibility of the issue being reported, and can follow any discussion on it.

Karql: Sorry I wanted to link these places together but forgot to past link to issue here.

Full details can be found here: https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/285

PR #580, opened 2022-04-13
Fix unit tests on Windows.

johnou:
iirc default charset on Windows is windows-1252 which causes unit tests to fail https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/569#issuecomment-1097698884 we can work around this by specifying utf8 in the gradle build file.


mc1arke:
Merged. Thanks for the quick fix!

PR #593, opened 2022-04-22
Add bitnami documentation section

ahonnecke:
Also lint the markdown

Signed-off-by: Ashton Von Honnecke <ashton@pixelstub.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ In case you want to try and test the current branch or build it for your develop
 # Using the bitnami AMI image
 
 ### Set up server from bitnami API
-- Spin up an EC2 instance with the bitnami AMI
+- Spin up an EC2 instance with the bitnami AMI (https://aws.amazon.com/marketplace/pp/prodview-zweq32he6zjmw?applicationId=AWSMPContessa#pdp-overview)
 - Set up SSL (process not documented here)
 - Set up DNS (process not documented here)
 - SSH into server (process not documented here)
--- a/README.md
+++ b/README.md
@@ -55,12 +55,12 @@ Set all other properties that you can define globally for all of your projects.
 In order to decorate your Pull Request's source branch, you need to analyze your target branch first.
 
 ### Run analysis of branches
-  
+
 The analysis needs the following setting:
 `sonar.branch.name = branch_name (e.g master)`
 
 ### Run analysis of the PR branch
-Carefully read the official SonarQube guide for [pull request decoration](https://docs.sonarqube.org/latest/analysis/pull-request/) 
+Carefully read the official SonarQube guide for [pull request decoration](https://docs.sonarqube.org/latest/analysis/pull-request/)
 
 In there you'll find the following properties that need to be set.
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,14 +72,39 @@ sonar.pullrequest.base = target_branch_name (e.g master)
...
+- Spin up an EC2 instance with the bitnami AMI (https://aws.amazon.com/marketplace/pp/prodview-zweq32he6zjmw?applicationId=AWSMPContessa#pdp-overview)
+- Set up SSL (process not documented here)
+- Set up DNS (process not documented here)
+- SSH into server (process not documented here)
+
+### Once SSHed into machine
+- `$ sudo su`
+- `$ cd /opt/bitnami/sonarqube/extensions/plugins/`
+- `$ wget https://github.com/mc1arke/sonarqube-community-branch-plugin/releases/download/1.11.0/sonarqube-community-branch-plugin-1.11.0.jar`
+-` $ chmod 777 sonarqube-community-branch-plugin-1.11.0.jar`
```
mc1arke, line 99: This opens up the file to being altered by any other user with access to the host, so is not an action we should be recommending to users

README.md
```diff
@@ -72,14 +72,39 @@ sonar.pullrequest.base = target_branch_name (e.g master)
...
+# Using the bitnami AMI image
+
+### Set up server from bitnami API
+- Spin up an EC2 instance with the bitnami AMI (https://aws.amazon.com/marketplace/pp/prodview-zweq32he6zjmw?applicationId=AWSMPContessa#pdp-overview)
+- Set up SSL (process not documented here)
+- Set up DNS (process not documented here)
+- SSH into server (process not documented here)
+
+### Once SSHed into machine
+- `$ sudo su`
```
mc1arke, line 96: We shouldn't immediately drop to root. If the actions are setting up files to be used by a processing running as the `bitnami` user then this should be `sudo su bitnami`

README.md
```diff
@@ -72,14 +72,39 @@ sonar.pullrequest.base = target_branch_name (e.g master)
...
+- `$ cd /opt/bitnami/sonarqube/extensions/plugins/`
+- `$ wget https://github.com/mc1arke/sonarqube-community-branch-plugin/releases/download/1.11.0/sonarqube-community-branch-plugin-1.11.0.jar`
+-` $ chmod 777 sonarqube-community-branch-plugin-1.11.0.jar`
+- `$ chown bitnami sonarqube-community-branch-plugin-1.11.0.jar`
+- `$ chgrp sonarqube sonarqube-community-branch-plugin-1.11.0.jar`
```
mc1arke, line 101: this could be simplifed to `chown bitnami:sonarqube sonarqube-community-branch-plugin-1.11.0.jar`, and shouldn't be necessary if we've `su`d to the right user at the start of the process.

README.md
```diff
@@ -72,14 +72,39 @@ sonar.pullrequest.base = target_branch_name (e.g master)
...
+- Set up SSL (process not documented here)
+- Set up DNS (process not documented here)
+- SSH into server (process not documented here)
+
+### Once SSHed into machine
+- `$ sudo su`
+- `$ cd /opt/bitnami/sonarqube/extensions/plugins/`
+- `$ wget https://github.com/mc1arke/sonarqube-community-branch-plugin/releases/download/1.11.0/sonarqube-community-branch-plugin-1.11.0.jar`
+-` $ chmod 777 sonarqube-community-branch-plugin-1.11.0.jar`
+- `$ chown bitnami sonarqube-community-branch-plugin-1.11.0.jar`
```
mc1arke, line 100: Does sonarqube not run under the `sonarqube` user in a bitnami distribution?

mc1arke (commented):
I'm not comfortable that this provides the right recommendations to users from a security perspective (overly broad file permissions), and is specific to a particular software packaging solution.

The actual issue seems to exist due to an attempt to perform an installation without checking Linux file ownership, which isn't something that this plugin's documentation should need to guide on.

ahonnecke:
>  I'm not comfortable that this provides the right recommendations to users from a security perspective (overly broad file permissions), and is specific to a particular software packaging solution.

> The actual issue seems to exist due to an attempt to perform an installation without checking Linux file ownership, which isn't something that this plugin's documentation should need to guide on.

I'm unclear on the feedback here, are you suggesting that I close this PR, or that I fix this PR?

mc1arke:
I'd probably suggest you close this PR. It's not reasonable to try and cover off edge cases for specific system installations in a readme, and I'd expect  someone using Linux to check file permissions and ownership as part of any review if they did discover an issue.

ahonnecke:
> I'd probably suggest you close this PR.

:+1: 

For future reference to anyone who stumbles upon this, this might save you some time:

- The documentation does not currently specify what ownership and permissions are required for the plugin to work, if you find that it's failing inexplicably you might be served by playing with the permissions; my experience is that `664` will fail with "Current edition does not support branch feature", I think maybe `700` is what you want, but I have not checked.
- The documentation currently references a `config` directory as the location of the `sonar.properties` file; my experience with the `mc1arke/sonarqube-with-community-branch-plugin` docker image, and the bitnami AMI from VMware is that the folder is actually named `conf`.
- The plugin appearing in the list of installed plugins in the web console does not necessarily indicate that it is installed and working.

PR #594, opened 2022-04-24
Add path correction for docker, lint markdown

ahonnecke:
Signed-off-by: Ashton Von Honnecke <ashton@pixelstub.com>


mc1arke:
Thanks for the contribution, but the equivalent change has been applied to master under a different Pull Request

PR #626, opened 2022-06-18
Add instructions for official SonarQube Helm Chart

rgembalik:
# Instructions for Helm Chart users

## What it does

This PR adds instructions for the plugin installation for [Sonarqube official Helm Chart](https://github.com/SonarSource/helm-chart-sonarqube/tree/master/charts/sonarqube) users.

The instruction also covers problems with incorrect copy paths, which are occurring only when persistence is enabled in installation.

## Why is it needed?

The plugin works well with a kubernetes environment, however, this use case is not covered in installation instructions, which are not trivial.

_Related to #609_


mc1arke:
Merged. Many thanks for the contribution!

PR #637, opened 2022-07-01
Don't fail GitLab pipeline on QualityGate failure - optional property

### LINKED ISSUE #137: [Gitlab] Switch to turn off failing of pipeline
**Describe the Enhancement**
Switch to turn off failing of pipeline

**Additional context**
If quality gate is not reached it is not possible to merge, which is to hard for us

szpak:
This PR adds an extra property `com.github.mc1arke.sonarqube.plugin.branch.pullrequest.gitlab.dontFailPipeline` (disabled by default) which if set to "true" do not fail GitLab pipeline on its status update.

As explained in #137, it might be useful in the interim period and while given changes cannot be feasibly tested (and we don't want to have the merge option blocked).

To simplify testing, I added `junit-jupiter-params` and migrated `GitlabMergeRequestDecoratorTest` to JUnit Jupiter. Nevertheless, if you don't like that change, it is in a separate commit and the implementation is done just in the first commit (I can remove the others, if requested).

I'm not sure about the name. Any better proposal is welcome. It would be also useful to put it somewhere in README, I haven't found a good section for that :-/.

Feel free to write if anything here could be improved.

Closes #137.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
@@ -52,6 +52,8 @@ public class GitlabMergeRequestDecorator extends DiscussionAwarePullRequestDecor
     public static final String PULLREQUEST_GITLAB_PROJECT_URL = "sonar.pullrequest.gitlab.projectUrl";
     public static final String PULLREQUEST_GITLAB_PIPELINE_ID =
             "com.github.mc1arke.sonarqube.plugin.branch.pullrequest.gitlab.pipelineId";
+    public static final String PULLREQUEST_GITLAB_DONT_FAIL_PIPELINE =
+            "com.github.mc1arke.sonarqube.plugin.branch.pullrequest.gitlab.dontFailPipeline";
 
     private final GitlabClientFactory gitlabClientFactory;
     private final MarkdownFormatterFactory formatterFactory;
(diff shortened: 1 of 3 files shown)


szpak:
Playing with the change on a local Sonar instance it turned out that only properties starting with `sonar.analysis.` as available from the analysis context. I changed the name accordingly.

Btw, `PULLREQUEST_GITLAB_PIPELINE_ID` is also not available for the plugin. I leave it as is as it is not related to my change.

```
    public static final String PULLREQUEST_GITLAB_PIPELINE_ID =
            "com.github.mc1arke.sonarqube.plugin.branch.pullrequest.gitlab.pipelineId";
```

src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
```diff
@@ -121,11 +123,15 @@ protected void submitPipelineStatus(GitlabClient gitlabClient, MergeRequest merg
         Long pipelineId = analysis.getScannerProperty(PULLREQUEST_GITLAB_PIPELINE_ID)
                 .map(Long::parseLong)
                 .orElse(null);
+        boolean dontFailPipeline = analysis.getScannerProperty(PULLREQUEST_GITLAB_DONT_FAIL_PIPELINE)
+                .map(Boolean::parseBoolean)
+                .orElse(false);
```
mc1arke, line 128: This is a facility we don't really want: someone who has access to submit scans can decide whether the pipeline passes on a Quality Gate failure, rather than it being clear that a passing pipeline is because of clean code. In the companies I work with, the decision is made up-front as to what rules must be adhered to, and it is assumed that a pipeline is passing because those rules have been met. If we start allowing people who control the scanner to decide whether they can bypass Quality Checks then we'd effectively be breaking the safeguard that Sonarqube is providing in these scenarios.

PR #660, opened 2022-08-18
Update README.md

Nimble85:
Update ${version} on Manual Install in order to avoid confusion.


mc1arke:
Could you add something along the lines of 'where `${version}` is the version of the plugin being worked with' to ensure we don't just end up with reports from people who copy and paste the example without understanding it?

Nimble85:
@mc1arke
Yes, I will add some comments near the `${version}` parameter. Thanks for your attention.

mc1arke:

PR #696, opened 2022-12-07
Support Pagination over AppInstallation

bstick12:
* Support pagination over the AppInstallation objects returned from Github.
* Fixes issue #508


mc1arke:
Merged! Thanks for the contribution!

PR #698, opened 2022-12-08
Some doco update / helfull code change

yaverhussain:


mc1arke (changes requested):
Your client change needs a unit test to protect it during any future work. I'd also suggest you split this into a PR per-commit since your 2 changes are doing very different things.

mc1arke:
Closing as abandoned.

Issue #674, opened 2022-10-18
Plugin v1.12.0 incompatible with SonarQube v9.7.0


stevehipwell:
**Describe the bug**
The v1.12.0 version of this plugin doesn't seem to be compatible with SonarQube v9.7.0.

**To Reproduce**
Install v1.12.0 to SonarQube v9.7.0.

**Expected behavior**
I'd expect the plugin to load as it does for v9.6.1.

**Screenshots**
> 2022.10.18 09:01:53 INFO  app[][o.s.a.ProcessLauncherImpl] Launch process[WEB_SERVER] from [/opt/sonarqube]: /usr/lib/jvm/java-11-openjdk/bin/java -Djava.awt.headless=true -Dfile.encoding=UTF-8 -Djava.io.tmpdir=/opt/sonarqube/temp -XX:-OmitStackTraceInFastThrow --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED --add-exports=java.base/jdk.internal.ref=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED --add-opens=java.management/sun.management=ALL-UNNAMED --add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED -Dcom.redhat.fips=false -Xmx512m -Xms128m -XX:+HeapDumpOnOutOfMemoryError -javaagent:./extensions/plugins/sonarqube-community-branch-plugin.jar=web -Dhttp.nonProxyHosts=localhost|127.*|[::1] -cp ./lib/sonar-application-9.7.0.61563.jar:/opt/sonarqube/lib/jdbc/h2/h2-2.1.214.jar org.sonar.server.app.WebServer /opt/sonarqube/temp/sq-process11500539336540629664properties
09:01:53.624 [main] INFO com.github.mc1arke.sonarqube.plugin.CommunityBranchAgent - Loading agent
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #699, opened 2022-12-10
#674: Add support for Sonarqube 9.7

mc1arke:
Sonarqube has removed the use of the same external component key and differing database IDs for different branches on the same projects and now use different IDs in all cases. The pull request web service endpoints have also been removed from community edition. To allow the plugin to work with the new version of Sonarqube, the component key generation for branches has been modified to save branch DTOs whenever a new branch is created, and to remove the conditions around re-using the same branch if the target branch details matched an existing branch. The Pull Request endpoint actions have been copied from the old community edition sources, and tidied up to use a cleaner abstraction model. As the front-end only shows branch features if the implementation of `BranchFeatureExtension` returns the name 'branch-support', the `CommunityBranchFeatureExtension` has been altered to follow this requirement, and an additional `MonoRepoFeature` has been implemented to allow the mono-repo switches to be shown against the front-end.

Includes the migration of any altered unit tests to JUnit 5.


MCMicS (approved):

RedlineTriad:
@mc1arke When do you plan to create a release with these changes?
Would be nice to be able to update Sonar to latest.

Issue #790, opened 2023-08-03
[Gitlab] Edit Summary Note instead of creating a new one every time


alexintech:
**Problem**
When we rerun a pipeline the plugin resolves old summary note and creates a new one, even if nothing changed and no new issue found. That clutters MR notes.

**What I'd like to see**
I suppose that editing existing Summary Note (even if it had been resolved earlier) will produce fewer mess.

j-martin:
@alexintech I see you got your own fork with this feature implemented. How close were you to opening a PR against the main repo? We'd love to have this feature on our sonarqube instance.

alexintech:
> @alexintech I see you got your own fork with this feature implemented. How close were you to opening a PR against the main repo? We'd love to have this feature on our sonarqube instance.

@j-martin PR has been closed by maintainer.  More information you can find [here](https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/791)

mc1arke:
I've raised #997 with an alternative approach which deleted comments with no follow-on threads, rather than editing in-place comments

PR #791, opened 2023-08-03
#790: Add an ability to edit Summary Note instead of creating a new one

alexintech:
Closes https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/790

* Added configuration option to allow edit existing Summary Note instead of creating a new thread;
* Added tests for the current behavior;
* GitlabMergeRequestDecoratorTest -> switch to JUnit 5;
* Some efforts to simplify creation of mocks for GitlabMergeRequestDecoratorTest and minimize duplication.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/CommunityBranchPlugin.java
@@ -73,6 +73,7 @@ import org.sonar.core.extension.CoreExtension;
 public class CommunityBranchPlugin implements Plugin, CoreExtension {
 
     public static final String IMAGE_URL_BASE = "com.github.mc1arke.sonarqube.plugin.branch.image-url-base";
+    public static final String PR_SUMMARY_NOTE_EDIT = "com.github.mc1arke.sonarqube.plugin.branch.pullrequest.summary.edit";
 
     @Override
     public String getName() {
@@ -157,6 +158,15 @@ public class CommunityBranchPlugin implements Plugin, CoreExtension {
                                           .build(),
                 MonoRepoFeature.class);
 
+            context.addExtensions(PropertyDefinition.builder(PR_SUMMARY_NOTE_EDIT)
+                    .category(getName())
+                    .subCategory("Merge Request Decoration")
+                    .onQualifiers(Qualifiers.PROJECT)
+                    .name("Edit summary note")
+                    .description("Edit summary discussion thread instead of resolving it and creating a new one (Gitlab only).")
+                    .type(PropertyType.BOOLEAN)
+                    .defaultValue(String.valueOf(false))
+                    .build());
         }
     }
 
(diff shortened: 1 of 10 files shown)


mc1arke:
We used to edit summary notes but moved away from it for a few reasons:
1. Gitlab does not notify when a note is edited, so reviewers don't know if something has changed
2. No history is recorded in an edit, so any conversation that spawns from an item in the summary note may be misleading given the final summary note content
3. Allowing configuration for this is non-standard so isn't covered under the Sonarqube documentation, and is therefore difficult to point people at

I'm presuming the driver behind this is that Gitlab doesn't collapse summary notes when the thread is resolved, so would suggest an alternative approach of creating a thread to add summary comments to with the initial message being `Sonarqube Scan Result` or something similar so that the resolved thread collapses to a small message

alexintech:
Sorry for the late response.

> 1. Gitlab does not notify when a note is edited, so reviewers don't know if something has changed

Yes, it's true. But Gitlab notifies if all issues have been resolved.

Expect another situation. Developer opens MR, SonarQube summary note is OK. Then developer pushes new commits, and every time SonarQube is OK. It sends a lot of annoying notifications. 

> 2. No history is recorded in an edit, so any conversation that spawns from an item in the summary note may be misleading given the final summary note content

I don't know how exactly SonarQube Developer Edition decorates MR's. All I have found is [this video](https://www.youtube.com/watch?v=u2dvQx08nXM). As I understand it updates existing summary note, and does not creates a new one.

Other tools also usually edit summary notes in Gitlab's MR. Gitlab's junit reports, code climate reports in MR also shows only the current state.

Also in our experience conversations usually spawns on individual line comments and never on summary note.

I presume that somebody should have a history for any change, so I added the configuration option to save the same behavior.

> 3. Allowing configuration for this is non-standard so isn't covered under the Sonarqube documentation, and is therefore difficult to point people at

As I understand from the [mentioned video](https://www.youtube.com/watch?v=u2dvQx08nXM), current behavior is already non-standard.

Many forks from this repository implements similar options, but that forks are not very popular or are not supported in long term. I suppose that making configuration options for some different behaviors would satisfy community needs.

Some additional documentation can be added to `README` or can be published some static docs on github pages.

> I'm presuming the driver behind this is that Gitlab doesn't collapse summary notes when the thread is resolved, so would suggest an alternative approach of creating a thread to add summary comments to with the initial message being Sonarqube Scan Result or something similar so that the resolved thread collapses to a small message

Not only collapsing, also annoying notifications. I suppose that in your approach viewing the latest (actual) summary would not be convenient.

mc1arke:
The linked video doesn't show a re-scan of a Merge Request, only the manual change of the status of a finding and a (realtime?) submission of a change in status to Gitlab. The video appears to show that the existing comment is deleted and a new one submitted given the lack of `edited` marker and the timestamp of the summary comment, which would be a preferred option given the new message would result in a new notification being generated for anyone watching the MR.

I also don't want us adding more configuration flags into the plugin unless there's a strict requirement to need a flag (i.e. the plugin won't work without setting a certain value in a configuration flag). The plugin takes an opinionated approach to operation to save on the overhead of testing combinations of configuration or handing different set-ups when investigating issues.

OneCyrus:
I really want to this feature as well. currenlty a lot of the PRs look like the screenshot below in GitHub for us. It would be great if we could reduce the noise by updating an existing comment. 
also another issue is that it doesn't work properly in monorepos as it doesn't respect project keys and closes everything even from other keys.

image

hballangan-mdsol:
To be honest, I've been pushin sonarqube in our CI and this has been one of the reviews of users as well. It creates too much noise since it notifies everyone for every commit. Ideally we only want to get emails if it's a user review, not for PR decorations

Issue #759, opened 2023-04-20
Please don't block the flow(just ignore the PR which missing the pull_request_binary).


bbxfnet:
**Describe the bug**
```
java.lang.NullPointerException: Pull request data should be available for branch type PULL_REQUEST
	at java.base/java.util.Objects.requireNonNull(Unknown Source)
	at com.github.mc1arke.sonarqube.plugin.server.pullrequest.ws.pullrequest.action.ListAction.addPullRequest(ListAction.java:123)
	at com.github.mc1arke.sonarqube.plugin.server.pullrequest.ws.pullrequest.action.ListAction.lambda$handleProjectRequest$2(ListAction.java:102)
	at java.base/java.util.ArrayList.forEach(Unknown Source)
	at com.github.mc1arke.sonarqube.plugin.server.pullrequest.ws.pullrequest.action.ListAction.handleProjectRequest(ListAction.java:102)
	at com.github.mc1arke.sonarqube.plugin.server.pullrequest.ws.pullrequest.action.ProjectWsAction.handle(ProjectWsAction.java:61)
```

**To Reproduce**
Unknow the reason which the PR don't have pull_request_binary data

**Expected behavior**
Please don't block the flow(just ignore the PR which missing the pull_request_binary).

**Screenshots**
None

**Software Versions**
 - SonarQube Version: sonar-application-9.9.0.65466.jar
 - Plugin Version: sonarqube-community-branch-plugin-1.14.0.jar

**Additional context**
None

mc1arke:
The error you're hitting indicates that a pull request is recorded in the database, but that it has no data recorded against it. This is definitely invalid (you have a pull request with no source branch) so I'd need you to provide more information about what the data in the database looks like, and also any other information you can think of (i.e. is this a new project that auto-created in an invalid state or one that has existed with working branches and PRs until now, is this only impacting one project are multiple ones, do you actively do anything like deleting PRs or branches from Sonarqube etc).

I don't think that removing that check is necessarily the right direction to go in since it's highlighting that something elsewhere in this plugin - or in your wider Sonarqube instance - is generating invalid data, and this could lead to further issues for you elsewhere.

mc1arke:
Closing as not-planned given there's no replication steps on this, and the check if there to prevent Sonarqube doing things that may further damage a project that already contains invalid data.

saurabh-sp-tripathi:
[...]

PR #818, opened 2023-10-24
Let ListAction skip pull requests having no binary data

reda-alaoui:
Following https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/759#issuecomment-1777469548, this change lets `ListAction` skip pull requests without binary data.


mc1arke:
Thanks for the contribution. Is this effectively a duplicate of #796? One thing I'm still missing from this is replication steps: not having part of the data on a pull request seems like an invalid scenario we should be preventing elsewhere, rather than attempting to return a response that technically isn't valid for one of the APIs

mc1arke:
I'm superceding this with #847 since the Pull Request data fields are not mandatory and I'd rather we didn't end up with data in the database that wasn't visible in the API, as users then don't know if they've got potentially broken data needing cleaned-up. Thanks for the contribution though!

Issue #738, opened 2023-02-23
Projects created automatically are broken


felipecrs:
**Describe the bug**
If I run `mvn sonar:sonar` with the `sonar.pullrequest` parameters for a project that does not already exist in Sonarqube, the project gets created automatically.

However, in that case, the project that gets created in Sonarqube is apparently broken because when I try to open it from the UI this shows up:

![image](https://user-images.githubusercontent.com/29582865/221047370-37232253-451f-4712-80b2-0ffbc939529d.png)

Digging a bit further, this issue is apparently related to the branch name of the project. I think the project does not get created by the plugin with the correct branch name set as the main branch.

If I create the project manually first in SonarQube, and set its main branch to the correct branch, then `mvn sonar:sonar` works as expected.

**To Reproduce**
1. Run `mvn sonar:sonar` with the `sonar.pullrequest` parameters for a project that does not exist already in Sonarqube.
1. After the analysis runs, go to Sonarqube and try to open the project from its UI. It should fail with the same error as the screenshot of earlier.

**Expected behavior**
To have the project created with the target branch `sonar.pullrequest.base` set as the default branch.

**Screenshots**
See above.

**Software Versions**
 - SonarQube Version: 9.9 (build 65466)
 - Plugin Version: 1.14.0

**Additional context**
This causes the Sonar Gerrit Jenkins plugin to fail for some weird reasons like:

```console
SonarQube analysis 'AYaAYSId_x56GYGzmtJ8' failed with message: Could not find target branch 'null' in project.
```

The Sonarqube logs shows things like:

```
ERROR ce[AYaAYSId_x56GYGzmtJ8][o.s.c.t.s.ComputationStepExecutor] Execution of listener failed
java.lang.IllegalStateException: Branch has not been set
java.lang.IllegalStateException: Could not find target branch 'null' in project
ERROR web[AYZ/fJ+hFk6S4/ySAAUs][o.s.s.w.WebServiceEngine] Fail to process request http://<host>/sonarqube/api/project_pull_requests/list?project=<project>%3Aspring-app-parent
java.lang.NullPointerException: Pull request data should be available for branch type PULL_REQUEST
```

mc1arke:
[...]

PR #847, opened 2024-01-01
#738: Prevent missing pull request data from causing API errors

mc1arke:
The scanner currently skips validation of a target branch if a Pull Request is used to create a new project, so the resulting project fails to load in front-end due to the Pull Request API treating the data on that pull request as invalid. This is being overcome by validating that a target branch exists for all Pull Request submissions and rejecting the scan submission if the target branch is not found in Sonarqube.

Additionally, there's a delay between a Pull Request being recorded in the database by the server component as a result of the call from the scanner, and the Compute Engine recording the Pull Request details (source, target, title etc.) against the branch. During this time the Pull Request treats that Pull Request as invalid and throws an error, meaning the project cannot be loaded through the UI, or the Pull Requests listed through the API. As the Pull Request response fields filled from the Pull Request data are not mandatory, those fields are now only being completed if the Pull Request data is set on the branch DTO rather than throwing an exception if the data isn't set.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/scanner/BranchConfigurationFactory.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/scanner/BranchConfigurationFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Michael Clarke
+ * Copyright (C) 2022-2024 Michael Clarke
  *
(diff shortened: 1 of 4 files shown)


bstick12:
@mc1arke will this fixed be merged back to the the plugin version 1.14.x supporting the 9.9 LTS release

PR #864, opened 2024-01-28
Added default value for commentType area.

nadidurna:
With this contribution Azure DevOps can see comment as text. Without this, commentType field was sending as unknown. Therefore Azure DevOps did not deliver to "new comment" email to subscribers.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/almclient/azuredevops/model/Comment.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/almclient/azuredevops/model/Comment.java
@@ -19,7 +19,7 @@ public class Comment {
                    @JsonProperty("commentType") CommentType commentType) {
         this.content = content;
         this.author = author;
-        this.commentType = commentType;
+        this.commentType = CommentType.TEXT;
     }
 
     /**


src/main/java/com/github/mc1arke/sonarqube/plugin/almclient/azuredevops/model/Comment.java
```diff
@@ -19,7 +19,7 @@ public Comment(@JsonProperty("content") String content, @JsonProperty("author")
                    @JsonProperty("commentType") CommentType commentType) {
         this.content = content;
         this.author = author;
-        this.commentType = commentType;
+        this.commentType = CommentType.TEXT;
```
mc1arke, line 22: This doesn't seem right to me: we're ignoring the parameter passed into the constructor and hardcoding a value, which means we never use the value actually returned by the API. Additionally this class is only used on API responses - not requests - so shouldn't impact whether Azure DevOps sends emails or not.

I'd expect the CreateCommentRequest class to have a `commentType` field added.

mc1arke (changes requested):
Given you're reporting something isn't working, your change needs to include a unit or integration test that highlights the issue and validates the fix.

PR #884, opened 2024-03-06
GitHub: minimize only comments with same project key

OneCyrus:
this addresses the issue of minimizing all sonarqube comments in a monorepo with multiple project keys

#785

PR #898, opened 2024-04-17
Fix detection of previous PR annotation comments

GreyTeardrop:
# Description

It seems like the https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/884 change has affected minimization of comments for SonarQube instances that run on Linux or macOS.

The root cause is that `MarkdownFormatterFactory` uses `System.lineSeparator()` as the end-of-line character for annotation comments. `GraphqlGithubClient.postSummaryComment()`, however, uses hardcoded `\r\n` as end-of-line character, so it isn't able to match legitimate comments if SonarQube is not running on Windows.

This change updates `GraphqlGithubClient` to also use system end-of-line char.

(It also moves `String.format` call out of the comment iteration logic to avoid making multiple identical calls).


OneCyrus (approved):
looks good!

nklarman:
Thank you @GreyTeardrop , @mc1arke is it possible to merge this one soon please? it's very annoying to have spam of comments on PRs.

bgong-mdsol (approved):

this-is-tobi (approved):

mc1arke:

PR #1002, opened 2024-11-18
Update README.md

TimoHauer:


README.md
```diff
@@ -23,7 +23,7 @@ Use the following table to find the correct plugin version for each SonarQube ve
 
 | SonarQube Version | Plugin Version |
 |-------------------|----------------|
-| 10.6              | 1.22.0         |
+| 10.7 & 10.6       | 1.22.0         |
```
brother, line 26:
```suggestion
| 10.6+                 | 1.22.0         |
```

I prefer this solution and it seems to have been used in earlier instances where same plugin version applied to multiple SQ versions.

TimoHauer: Thank you for the feedback. I changed it. Later, I realized that I could have used your suggestion.

mc1arke:
Closing since this has been performed as part of a subsequent release cycle

PR #1003, opened 2024-11-18
fix: Update GitHub PR check name

### LINKED ISSUE #415: [GitHub] Make pull request status check name configurable
Hi!

Following idea for an enhancement: it would be great if the name of the status check on a pull request would be configurable. At the moment it is hard-coded to `<project-name> Sonarqube Results`, see [the source code](https://github.com/mc1arke/sonarqube-community-branch-plugin/blob/1.9.0/src/main/java/com/github/mc1arke/sonarqube/plugin/almclient/github/v4/GraphqlGithubClient.java#L127).

Which leads to a status check like this (where `sonarqube-test` is the project name):

image

This makes it hard to apply global policies for required status checks in GitHub, because the name differs per project. As an example: there is a global policy we apply to all GitHub repositories that a status check called "Jenkins" is required to pass before a PR could be merged. That is no problem, because the status check is called "Jenkins" for every repository.

If I would like to do the same for SonarQube, I would need to adjust the name of the status check in every repository, e.g.:

image

If I could support by testing a new plugin version or similar, I would be happy to do so.

jackylamhk:
Fixes #415

Use a static check name instead, which makes it easier to apply global rulesets to repositories.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/GithubPullRequestDecorator.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/GithubPullRequestDecorator.java
@@ -51,6 +51,7 @@ import com.github.mc1arke.sonarqube.plugin.ce.pullrequest.report.ReportGenerator
 
 public class GithubPullRequestDecorator implements PullRequestBuildStatusDecorator {
 
+    private static final String DEFAULT_CHECK_RUN_NAME = "SonarQube Code Analysis";
     private final GithubClientFactory githubClientFactory;
     private final ReportGenerator reportGenerator;
     private final MarkdownFormatterFactory markdownFormatterFactory;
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/github/GithubPullRequestDecorator.java
```diff
@@ -102,7 +102,7 @@ private GHPullRequest createCheckRun(GHRepository repository, AnalysisDetails an
             );
         }
 
-        repository.createCheckRun(String.format("%s Sonarqube Results", analysisDetails.getAnalysisProjectName()), analysisDetails.getCommitSha())
+        repository.createCheckRun("SonarQube Code Analysis", analysisDetails.getCommitSha())
```
mc1arke, line 105: I think we need to cater for the scenario where a Monorepo is being used, so have some logic in here that includes the project name when Monorepo is set?

mc1arke:

Issue #1078, opened 2025-06-05
Dropdown list disabled after upgrade to 25.6


SuhovDE:
After upgrading SonarQube server from **v25.5** to **v25.6**, the branch selection dropdown in the plugin UI stopped working. While the plugin worked flawlessly with v25.5, the dropdown no longer shows available branches or reacts to user input in v25.6.

### Affected Plugin Version

- **Plugin Version**: 1.24 (release)

### Affected SonarQube Version

- **Before Upgrade (Working)**: 25.5
- **After Upgrade (Broken)**: 25.6

### Expected Behavior

Branch dropdown should list all available branches for the current project and allow users to switch between them directly in the UI.

### Actual Behavior

- The dropdown is disable again.
- It is impossible to switch branches using the dropdown UI.
- However, navigation via:
  - **Background Tasks** page,
  - or manually appending `?branch=<branch_name>` to the URL,
  still works, which confirms that the branches exist and are accessible.

### Steps to Reproduce

1. Install plugin v1.24 on SonarQube 25.6.
2. Open a project dashboard.
3. Try to use the branch selection dropdown.

### Workarounds

- Use the **Background Tasks** page to navigate to branches.
- Manually append `?branch=<branch_name>` to the URL.

riteshkumar-prive:
I'm observing the same issue but with version 25.5 and artefacts from 1.24 release.

Sonarqube version: v25.5.0.107428

Observations:
- The branch drop-down UI component was working as expected with UI artefacts from [PR](https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/1067) 
- I removed the `web` directory and re-deployed the artefacts from [1.24](https://github.com/mc1arke/sonarqube-community-branch-plugin/releases/download/1.24.0/sonarqube-webapp.zip) release artefacts and it stopped working.

GEownt:
I have the same problem but with other versions.
I am using the image provided by mc1arke `mc1arke/sonarqube-with-community-branch-plugin:25.4.0.105899-community`
I was running `sonarqube:25.4.0.105899-community` before and changing to mc1arke image doesn´t do anything to the dropdown, it still doesn´t work for me.
This is what my docker-compose.yml look like:
```
services:
  sonarqube:
    image: mc1arke/sonarqube-with-community-branch-plugin:25.4.0.105899-community
    restart: always
    environment:
      SONAR_JDBC_USERNAME: my_user
[...]

PR #1081, opened 2025-06-06
#1078: Ensure we're building the customised webapp

mc1arke:
Older versions of sonarqube-webapp have the additional components in an `addons` rather than `sq-server-addons` directory, so the build customisation isn't creating a modified front-end. The project setup has therefore been updated to use the path the webapp expects, and the branch components altered to use the right imports for the target version of Sonarqube. To simplify the plugin versioning, the plugin version has been altered to match the major and minor version of the Sonarqube distribution it's aiming to support.

### DIFF
--- a/.env
+++ b/.env
@@ -5,4 +5,4 @@ SONARQUBE_VERSION=25.4.0.105899-community
 DOCKERFILE=Dockerfile
 
 # The version of the plugin to include in the image
-PLUGIN_VERSION=1.25.0-SNAPSHOT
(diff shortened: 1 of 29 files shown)


mc1arke:
Running `docker compose up`
image

Plus unpacking the webapp-build.zip
```shell
$ ls sonarqube-webapp/js | grep Branch
ProjectBranchesApp-ByqUtMj1.js
ProjectBranchesApp-ByqUtMj1.js.map
```

Issue #1097, opened 2025-06-27
25.6.0 image posts extra “SonarQube” build status to Bitbucket PRs, blocking merges


devogosling:
<p><strong>Describe the bug</strong><br>
After upgrading from <strong>SonarQube 10.7 + community-branch plugin 1.22.0</strong> to the latest Docker image <strong><code inline="">mc1arke/sonarqube-with-community-branch-plugin:25.6.0.109173-community</code></strong>, each pull-request build now creates <strong>two Bitbucket builds</strong>:</p>
<ul>
<li>
<p>the expected Code-Insights report (<code inline="">com.sonarsource.sonarqube</code>)</p>
</li>
<li>
<p><strong>and</strong> an extra build-status line (“SonarQube”) shown under <strong>Builds</strong></p>
</li>
</ul>
<p>The new build status is counted by Bitbucket’s “Minimum successful builds” merge-check, so any failed quality gate blocks a PR even when Code-Insights requirements are already disabled.<br>
With 10.7 + 1.22.0 only the Code-Insights report was posted, so builds were unaffected.</p>
<hr>
<p><strong>To Reproduce</strong></p>
<ol>
<li>
<p>Open a pull request in <strong>Bitbucket Server</strong> that is analysed by SonarQube.</p>
</li>
<li>
<p>Run analysis (we use <strong>SonarScanner 6.2.1.4610</strong>).</p>
</li>
<li>
<p>In the PR open <strong>Builds</strong>.</p>
</li>
<li>
<p>Observe an additional line, “SonarQube”, mirroring the quality-gate status.</p>
</li>
</ol>
<p>If we disable the <em>Minimum successful builds</em> merge-check the PR merges fine, confirming the new build status is what blocks it.</p>
<hr>
<p><strong>Expected behavior</strong><br>
Only the Code-Insights report should be created <strong>or</strong> there should be a switch to disable the build-status.<br>
Teams that rely on <em>Minimum successful builds</em> to enforce CI jobs can no longer use that check without also blocking on SonarQube.</p>
<hr>
<p><strong>Screenshots</strong><br>

![Image](https://github.com/user-attachments/assets/4f26dcc8-9293-4f50-b2f7-9fbec5d32744)

</p>
<hr>
<p><strong>Software Versions</strong></p>

Component | Version
-- | --
SonarQube | 25.6.0 (build 109173-community)
sonarqube-community-branch-plugin | 25.6.0
[...]

Issue #893, opened 2024-03-29
[Question]: should I expect to see SonarQube build for Bitbucket integrations


dbaggott:
**Describe the bug**

First and foremost, thank you for this valuable plugin.  Your work and contribution to the community is much appreciated!

I am not sure if this is a bug or expected behavior, so this is really a question as to what I should expect with regards to functionality when switching from the the Developer Edition to using the Community Edition Version 9.9.4 and version 1.14.0 of this lovely plugin.

We use Bitbucket for source control and, after making the transition from Developer to Community+plugin, the SonarQube builds that we were previously seeing in Bitbucket disappeared.  Fortunately, the code comments and the report are still present so the basic integration is working fine.

So, my question is: is there a way to get the SonarQube build back within Bitbucket or is that not supported with this plugin?

The screenshots below illustrate what I'm referring to if it's not clear.

**Bitbucket Integration with Developer Edition Screenshot**

Note the presence of a SonarQube build that is either green or red depending on the results (green in this case):

![Zight 2024-03-29 at 12 04 50 PM](https://github.com/mc1arke/sonarqube-community-branch-plugin/assets/112677/c801a5ef-fdc1-47ca-ae3d-9387c9347111)

**Bitbucket Integration with Community Edition + Plugin Screenshot**

Note the absence of a SonarQube build:

![Zight 2024-03-29 at 12 05 41 PM](https://github.com/mc1arke/sonarqube-community-branch-plugin/assets/112677/138334b0-566a-43b1-bad1-6c67d8812944)

**Software Versions**
 - SonarQube Version: Community Edition Version 9.9.4 (build 87374)
 - Plugin Version: 1.14.0

Thank you!

PR #1168, opened 2025-11-18
feature: Publish bitbucket build-status only on bitbucket cloud

ThibautMarechal:
Hi,

This change adds the ability to skip the build-status publishing that currently blocks pull-request merges. The functionality is redundant when using Bitbucket’s native feature to block pull-requests based on a Code Insight condition (see screenshot below).
image

This PR addresses the following issue: https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/1097.

I’m not a Java developer, so there may be a more elegant way to implement this. I also didn’t find any existing configuration in the repository that I could replicate for this functionality. 
Any guidance or suggestions for improvement would be greatly appreciated.

### DIFF
--- a/src/test/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/bitbucket/BitbucketPullRequestDecoratorTest.java
+++ b/src/test/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/bitbucket/BitbucketPullRequestDecoratorTest.java
@@ -45,6 +45,7 @@ import org.sonar.api.issue.impact.Severity;
 import org.sonar.api.issue.impact.SoftwareQuality;
 import org.sonar.ce.task.projectanalysis.component.Component;
 import org.sonar.ce.task.projectanalysis.component.ReportAttributes;
+import org.sonar.db.alm.setting.ALM;
 import org.sonar.db.alm.setting.AlmSettingDto;
 import org.sonar.db.alm.setting.ProjectAlmSettingDto;
 
@@ -132,6 +133,7 @@ class BitbucketPullRequestDecoratorTest {
     @Test
     void testNullPercentagesReplacedWithZeroValues() throws IOException {
         when(client.supportsCodeInsights()).thenReturn(true);
+        when(almSettingDto.getAlm()).thenReturn(ALM.BITBUCKET_CLOUD);
         AnnotationUploadLimit uploadLimit = new AnnotationUploadLimit(1000, 1000);
         when(client.getAnnotationUploadLimit()).thenReturn(uploadLimit);
 


dkeitzel:
I agree on the sentiment that the native 'Code Insights' mechanism should be used to indicate the status of a quality gate and whether it should block merging. This isn't in line with other ALM integrations, but is in line with the official SonarQube documentation, as @ThibautMarechal pointed out.

Well, at least for the Server edition of Bitbucket. As it turns out the behavior is different between the Server edition and the Cloud edition. Bitbucket Server uses [Code Insights](https://docs.sonarsource.com/sonarqube-server/devops-platform-integration/bitbucket-integration/bitbucket-server-integration/project#preventing-pull-request-merges-when-the-quality-gate-fails), Bitbucket Cloud uses [failed builds](https://docs.sonarsource.com/sonarqube-server/devops-platform-integration/bitbucket-integration/bitbucket-cloud-integration/project#preventing-pull-request-merges-when-the-quality-gate-fails).

I assume the reporter of #893 was using the Cloud edition of Bitbucket. One could argue #998 introduced a regression for all Bitbucket Server users.

I propose implementing the quality gate handling depending on the actual ALM type. There is `ALM.BITBUCKET` and `ALM.BITBUCKET_CLOUD`.

Furthermore I suggest using "Code Insights" exclusively and not using failed builds at all for `ALM.BITBUCKET`. This is available since Bitbucket Server version 6.9 (year 2009) on wards, which should be sufficient.

What do you think?

adrianwarowny:
Good idea it may help to workaround my issue: https://github.com/mc1arke/sonarqube-community-branch-plugin/issues/1171

Lauri-Nomme:
+1 

Thanks for implementing this, 100% support making SQ -> BB build-status optional.
The PR #998 indeed introduced a regression for us across multiple teams and we're currently working this around with removing the Minimum successful builds Merge Check on BB side, configuring Required builds with the necessary build keys that should actually block merges and having the systems reporting those builds generate stable build keys.

So, quite the hassle, would be nice having this disabled for BB server (or configurable) on the SQ side to remove the workarounds and have the old behavior back.

ThibautMarechal:
Hi @mc1arke,
I’m looking into how we can restore code insights support and I see a few possible approaches:

- Introduce a new environment variable, although that would differ from the current SonarQube behavior.
- Detect the Bitbucket Server version (the code insights blocking feature appeared in 6.9) and fall back to build statuses if the version is older.
- Detect whether we’re on Bitbucket Server or Bitbucket Cloud and only publish build statuses for Cloud.

Do you have a preference among these, or see a better direction?

We had to disable code insights on our side after a recent upgrade, so I’d be really glad to help move toward something that works again for everyone.

Thanks,
Thibaut

Lauri-Nomme:
@mc1arke any concrete arguments against this improvement?

mc1arke:
> @mc1arke any concrete arguments against this improvement?

I've got no issue with the concept, but the build doesn't currently pass

intgr:
```
BitbucketPullRequestDecoratorTest > testNullPercentagesReplacedWithZeroValues() FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at BitbucketPullRequestDecoratorTest.java:156
```

ThibautMarechal:
@mc1arke I normally fixed the test 🙂

mc1arke:
Merged, thanks for the contirbution

PR #1188, opened 2026-01-25
Add support for Sonarqube 25.12

mc1arke:
Upgrades to Sonarqube 25.12 for both back-end and front-end. As
Sonarqube no longer exposes commons lang, the relevant imports have been
replaced with lang3 imports. The front-end has also been altered to
replace `useBranchesQuery` references with `useProjectBranchesQuery` and
update imports for relocated components.


sebastiaanspeck:
The only quick issue is see is:
image 
when visiting `/admin/settings`.

PR #1200, opened 2026-02-05
Add support for Sonarqube 26.2

mc1arke:
Bumps to the backend for Sonarqube 26.2, and a recent front-end commit to use a fairly new front-end version. As the HttpClient version bundled with Sonarqube is not compatible with the Wiremock version used in tests, the Wiremock version has been specifically set in a scope that prepends it to the test runtime classpath to ensure its dependencies are used before any classes bundled with Sonarqube.

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 # The Sonarqube base image. 'latest' if building locally, '8.5-community' if targeting a specific version
-SONARQUBE_VERSION=26.1.0.118079-community
+SONARQUBE_VERSION=26.2.0.119303-community
 
 # The name of the Dockerfile to run. 'Dockerfile' is building locally, 'release.Dockerfile' if building the release image
 DOCKERFILE=Dockerfile
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ repositories {
     }
 }
 
-def sonarqubeVersion = '26.1.0.118079'
+def sonarqubeVersion = '26.2.0.119303'
 def sonarqubeLibDir = "${projectDir}/sonarqube-lib"
 def sonarLibraries = "${sonarqubeLibDir}/sonarqube-${sonarqubeVersion}/lib"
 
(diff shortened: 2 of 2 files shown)


hostalp:
Currently, your sonarqube-webapp directory points to commit https://github.com/SonarSource/sonarqube-webapp/commit/462b26884a759d830fede1755821b44c97263034 which isn't tagged `sqcb-26.2.0` - and there's currently no such tag at all.
Something similar happened already in January - the sqcb-26.1.0 tag was missing for some time (perhaps 2 weeks) until it appeared in that repository along with several commits.

Therefore I expect this to happen again this time - so you can either build the webapp from what's currently available and perhaps later rebuild it again, or wait for that sonarqube-webapp repository state to be fixed.
The current state likely works, but doesn't fully reflect the actual webapp in their 26.2 binary distribution yet. 

It seems like their GitHub repository is a mirror of their internal "master" repository and the synchronization doesn't work reliably. 

By the way, when looking at the code changes and comparing them with the built 26.2 web app, it looks like this is the last commit which is included in the built web app:
https://github.com/SonarSource/sonarqube-webapp/commit/ca8cf02b515a48f05cf9e9a5e2236c01b0af5839

Perhaps this one as well, but as it only contains ESLint rule changes it's hard to say without trying to compile the web app: https://github.com/SonarSource/sonarqube-webapp/commit/ff3470222bd3c3993608c56874d51d1354d649bd

The commit that follows it doesn't appear to be included (and none of those that follow it either): https://github.com/SonarSource/sonarqube-webapp/commit/0129fa6935470141805699035e7bef7452b63bd6

mc1arke:
Yeah, the lack of tag if why the commit currently uses a 'fairly new' front-end version rather than the actual tag. It's really to prove the front-end should build rather than being fully release ready. I'll keep an eye out for an appropriate tag being pushed.

kzhuklinets:
@mc1arke, thank you! A release creation has been skipped for this PR, is it intentional?

PR #1214, opened 2026-03-07
Fix authorization bypass in ProjectWsAction and DeleteAction

henricook:
👋🏻 Hello! I was just about to add this awesome looking project to my SonarQube Community install (thanks for your work!) and noticed a couple of issues on the security scan. Please let me know if I've missed a contributor step or, of course, if it's totally wrong and I should just close this and we can all forget I was ever here.

Issue: `hasEntityPermission()` returns a boolean that was silently discarded, meaning permission checks were never enforced. Replaced here with `checkEntityPermission()` which throws on failure.


mc1arke:

PR #1240, opened 2026-04-15
Fix typo in messages

sebastiaanspeck:


sebastiaanspeck:
I also want to mention that this message is reported on every run:
image

One-time commenting seems enough :wink:

mc1arke:
> One-time commenting seems enough 😉

There was supposed to be a check that prevents this being reported on every run (i.e treat a thread with this comment as already closed). I'll check what happened to that.

PR #1241, opened 2026-04-15
Prevent summary comment closure note being posted multiple times

mc1arke:
The filtering of open threads only treated a thread as closed if it wasn't a summary comment, so reposted the notice about a thread needing manually resolved on each execution of the decorator. The filtering of comments has therefore been changed to only consider whether a thread is resolved or has the manual resolution comment on it, without considering whether the thread is for a summary comment.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/DiscussionAwarePullRequestDecorator.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/DiscussionAwarePullRequestDecorator.java
@@ -191,19 +191,12 @@ public abstract class DiscussionAwarePullRequestDecorator<C, P, U, D, N> impleme
                     return commentsForDiscussion.stream()
                         .findFirst()
                         .filter(note -> isNoteFromCurrentUser(note, currentUser))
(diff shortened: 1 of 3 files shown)


benoit-waldmann:
This change appears to have introduced a regression for GitLab summary comments (observed on 26.5.0).

Removing the `|| isSummaryComment(...)` clause from `findOpenSonarqubeComments` means resolved threads are now uniformly excluded from the cleanup scan. The problem is that GitlabMergeRequestDecorator.submitSummaryNote resolves the summary discussion itself whenever the Quality Gate passes:

```
if (analysis.getQualityGateStatus() == QualityGate.Status.OK) {
    client.resolveMergeRequestDiscussion(...);
}
```

So on the next analysis, the previous summary is already resolved → it's filtered out by `!isResolved(...)` → it never reaches `deleteOrPlaceFinalCommentOnDiscussion` → it is never deleted. The result is that a new summary comment is added on every passing analysis and the old ones pile up. The `isSummaryComment `exception added in [#997](https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/997) was specifically guarding against this.

When the Quality Gate fails, the summary stays unresolved and cleanup still works correctly — which is why the bug only shows up on green MRs.

PR #1247, opened 2026-04-29
Expose a parameter to allow configuring inline annotations for GitLab

vincentscode:
After our GitLab server tried to send me about 400 e-mails for a single merge request, I wanted to be able to configure inline comments as it is possible for Azure.
The implementation should also be equivalent to the Azure one.

Unfortunately I have not found an easy way to expose this to UI. If there is one, I am happy to implement it!

It can be set via the API however, which I think is better than nothing.

### DIFF
--- a/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
+++ b/src/main/java/com/github/mc1arke/sonarqube/plugin/ce/pullrequest/gitlab/GitlabMergeRequestDecorator.java
@@ -44,6 +44,7 @@ import org.sonar.db.alm.setting.ProjectAlmSettingDto;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
 public class GitlabMergeRequestDecorator extends DiscussionAwarePullRequestDecorator<GitlabClient, MergeRequest, User, Discussion, Note> {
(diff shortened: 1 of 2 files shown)


mc1arke:
Thanks for the proposed contribution, but we match the upstream API definition for the plugin's implementation (https://next.sonarqube.com/sonarqube/web_api/api/alm_settings/set_gitlab_binding) which doesn't contain this parameter. Adding non-standard parameters causes problems for the future where Sonarsource changes the API, as we potentially end up with API conflicts, incompatible persisted data, or features that can't be retained from release-to-release.

If your concern is the number of emails you're receiving then I'd suggest looking at changing the decorator implementation to create draft notes and then submit them as a single batch (https://docs.gitlab.com/api/draft_notes/)

vincentscode:
Thank you for the suggestion, unfortunately that will not entirely fix the issue, as (at least our GitLab instance) is also severely unhappy with 400+ notes on a merge request in general and having that amount of spam is simply never helpful even if it were performant enough.

I do certainly understand your stance on matching upstream though! I hoped that sticking with an implementation that exactly matches the one of another backend might be fine as it is the most unlikely to cause issues.

Do you think there is any benefit to suggesting such an option upstream or would you recommend to simply keep maintaining my own fork with this small patch?

mc1arke:
I've not used the Gitlab decorator in Sonarqube commercial edition for a long time so don't know how it operates. If it doesn't create a comment per-issue then there's a reasonable debate for looking at altering how this integration runs, or if it does generate individual comments then it would need a check to see if Sonarqube has a way of disabling that, or raising a feature request with Sonarsource to allow that option to be disabled.

PR #1280, opened 2026-07-22
Add support for Sonarqube 26.7

### LINKED ISSUE #1277: ETA for plugin version 26.7 ?
Hi, just wanted to ask if there is an ETA for support of the latest sonarqube community edition (26.7) ?
I'm asking this because for sonarqube community edition 26.7 with plugin 26.5 we still have this issue https://community.sonarsource.com/t/cannot-change-quality-profile-of-a-project/182033 
Thank you

gsaraiva2109:
Closes #1277

Adds support for SonarQube Community Build 26.7 (26.7.0.124771).

Bumped the `sonarqubeVersion` in build.gradle, the .env defaults, and gradle.properties dev version. Skipped 26.6 entirely since no plugin release ever went out for it, so this jumps straight from 26.5.1 to 26.7.

One thing worth flagging for review: SonarSource hasn't tagged a 26.7 webapp snapshot yet (no `sqcb-26.7.0`), and 26.7's release notes are backend-only anyway, so I pinned the `sonarqube-webapp` submodule to `sqcb-26.6.0` instead of leaving it stale or pointing at `master` (which is tracking their internal 2026.x versioning and isn't compatible here).

Also had to fix `ScannerPullRequestPropertySensorTest`, it wouldn't compile anymore. `SensorContextTester` and `TestInputFileBuilder` got removed from the server distribution somewhere between 26.5 and 26.7. Rewrote it against a mocked `SensorContext` instead, same assertions, no changes needed on the production side.

Tested locally end to end: `./gradlew clean build` and the full test suite pass, and I ran it against a live 26.7 instance via docker-compose, plugin loads clean on both web and CE, and branch/PR analysis worked in my testing.

Happy to re-point the webapp submodule once SonarSource ships a 26.7 tag, if that's preferred over merging with the 26.6 one.

### DIFF
--- a/.env
+++ b/.env
@@ -1,8 +1,8 @@
 # The Sonarqube base image. 'latest' if building locally, '8.5-community' if targeting a specific version
-SONARQUBE_VERSION=26.5.0.122743-community
+SONARQUBE_VERSION=26.7.0.124771-community
 
 # The name of the Dockerfile to run. 'Dockerfile' is building locally, 'release.Dockerfile' if building the release image
 DOCKERFILE=Dockerfile
 
 # The version of the plugin to include in the image
-PLUGIN_VERSION=26.6.0-SNAPSHOT
+PLUGIN_VERSION=26.7.0-SNAPSHOT
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ repositories {
     }
 }
 
-def sonarqubeVersion = '26.5.0.122743'
+def sonarqubeVersion = '26.7.0.124771'
 def sonarqubeLibDir = "${projectDir}/sonarqube-lib"
 def sonarLibraries = "${sonarqubeLibDir}/sonarqube-${sonarqubeVersion}/lib"
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,1 @@
-version=26.6.0-SNAPSHOT
+version=26.7.0-SNAPSHOT
--- a/sonarqube-webapp-addons/src/branches/app/components/BranchPurgeSetting.tsx
+++ b/sonarqube-webapp-addons/src/branches/app/components/BranchPurgeSetting.tsx
@@ -18,7 +18,8 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
-import { HelperHintIcon, Spinner } from '~design-system';
+import { Spinner } from '@sonarsource/echoes-react';
+import { HelperHintIcon } from '~design-system';
 import { Switch } from '~adapters/components/common/Switch';
 import { translate } from '~sq-server-commons/helpers/l10n';
 import { useExcludeFromPurgeMutation } from '~sq-server-commons/queries/branch';
@@ -51,7 +52,7 @@ export default function BranchPurgeSetting(props: Props) {
         onChange={handleOnChange}
         value={branch.excludedFromPurge}
       />
-      <Spinner loading={isPending} className="sw-ml-1" />
+      <Spinner isLoading={isPending} className="sw-ml-1" />
       {isTheMainBranch && (
         <HelpTooltip
           className="sw-ml-1"
--- a/sonarqube-webapp-addons/src/branches/app/components/LifetimeInformationRenderer.tsx
+++ b/sonarqube-webapp-addons/src/branches/app/components/LifetimeInformationRenderer.tsx
@@ -18,9 +18,10 @@
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+import { Spinner } from '@sonarsource/echoes-react';
 import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
-import { Link, Spinner } from '~design-system';
+import { Link } from '~design-system';
 import { translate } from '~sq-server-commons/helpers/l10n';
 import { formatMeasure } from '~sq-server-commons/sonar-aligned/helpers/measures';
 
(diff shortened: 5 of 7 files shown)


Magyarimiki:
hi,

to fix the QP selection issue, you need this commit from webapp. This is not included in 26.6.0 version of webapp as it has been committed about a week later.

    commit 2ee7b79739d7c46be58fced5ad1bf2ba3c270ab0
    Author: David Cho-Lerat <david.cho-lerat@sonarsource.com>
    Date:   Mon May 18 10:00:16 2026 +0200

    SONAR-27859 Fix quality profile select dropdown appearing below the modal overlay

    GitOrigin-RevId: 412aa81f9a42264cfdf765d1ec32adf72f2c6975

Magyarimiki:
hi,

webapp now has the sqcb-26.7.0 tag . I did a quick check and the above mentioned commit for the UI fix is included.
I guess you can use this tag for this PR and we are good to go.

gsaraiva2109:
> hi,
> 
> webapp now has the sqcb-26.7.0 tag . I did a quick check and the above mentioned commit for the UI fix is included. I guess you can use this tag for this PR and we are good to go.

Hi, I did update the webpack, now I'm just waiting for the author to code review and merge...

sentoz:
@mc1arke Could you please review this PR?

renatoaraujo:
FYI, in case it is useful: we needed 26.8 (26.8.0.126808) and it turned out to work with the same changes as this PR plus one extra bit — `source/targetCompatibility` has to go from 17 to 21, since 26.7+ jars are class file major 65 and compiling against them with 17 fails (`class file has wrong version 65.0, should be 61.0`).

With that, `gradle build` (including the test suite, with the `ScannerPullRequestPropertySensorTest` fix from this PR) is green against 26.8, and the plugin loads fine as a javaagent on both web and CE: `branch-support` shows up in `api/features/list`, and PR analysis + GitHub PR decoration work on a real project. Nothing in the plugin main sources needed touching.

Not asking for anything — just noting that the jump does not seem to require more than a version bump either. Thanks for the plugin.

rwa-pwx:
Heads-up: the `sonarqube-webapp` submodule pin on this branch doesn't build.

The PR description says the submodule was pinned to `sqcb-26.6.0`, but the current head
(`1355871`) has it at `817db09`, which is `sqcb-26.7.0` — presumably the re-point after
@Magyarimiki's suggestion above. `sq-server:build` fails on that tag:

```
error during build:
../../libs/sq-server-addons/src/branches/components/branch-list/BranchList.tsx (22:34):
"Spinner" is not exported by "../../libs/sq-server-commons/src/design-system/index.ts",
imported by "../../libs/sq-server-addons/src/branches/components/branch-list/BranchList.tsx".

 NX   Running target build for project sq-server failed
```

**Cause:** `sqcb-26.7.0` deletes `libs/sq-server-commons/src/design-system/components/Spinner.tsx`
and its barrel export.

| webapp tag | `design-system/components/Spinner.tsx` |
|---|---|
| `sqcb-26.6.0` | present (3534 bytes), exported at `components/index.ts:81` |
| `sqcb-26.7.0` | absent |
| `sqcb-26.8.0` | absent |

Three files in `sonarqube-webapp-addons` still import it from `~design-system`:

- `src/branches/app/components/BranchPurgeSetting.tsx:21` → `<Spinner loading={isPending} …>`
- `src/branches/app/components/LifetimeInformationRenderer.tsx:23` → `<Spinner loading={loading}>` (used as a wrapper, with children)
- `src/branches/components/branch-list/BranchList.tsx:22` → `<Spinner />`

Worth noting the addons library already uses a *different* Spinner elsewhere —
`BranchNewCodeDefinitionSettingModal.tsx` and `PullRequestOverview.tsx` import it from
`@sonarsource/echoes-react` and pass `isLoading` rather than `loading`. So this looks like a
partially-completed migration rather than a one-line import swap: the prop contract differs, and
`LifetimeInformationRenderer` uses it as a wrapper. Rollup also stops at the first unresolved
export, so there may be more behind this one.

**Why CI didn't catch it:** there are no check runs on `1355871` — the only status on the commit is
`security/snyk`. The `build-ui` job that would have caught this never ran on this branch, so the
green-looking PR isn't covering the frontend.

**Data point on 26.8, corroborating @renatoaraujo above:** we're running 26.8 (`26.8.0.126808`)
built from this branch's plugin sources with `sonarqubeVersion` bumped and
`source/targetCompatibility` moved 17 → 21, and it works — plugin loads on both web and CE,
`branch-support` present, branch analysis fine. The only change we needed beyond that was pinning
the webapp back to `sqcb-26.6.0`, since the addons library compiles against that untouched. So the
frontend is currently the constraint, not the plugin.

Not asking for anything — just flagging that the pin as it stands won't produce a working image.
Thanks for the plugin, and for the 26.7/26.8 work.

gsaraiva2109:
Good catch, thanks for the detailed writeup. You were right on all counts.

Confirmed it locally: Spinner.tsx is gone from design-system as of sqcb-26.7.0 (also checked 26.8.0, still gone there), and BranchList.tsx, BranchPurgeSetting.tsx and LifetimeInformationRenderer.tsx were still pulling it from ~design-system. Just pushed a fix that switches all three to the echoes-react Spinner instead, same one BranchNewCodeDefinitionSettingModal and PullRequestOverview already use, just had to rename the loading prop to isLoading since that's what echoes-react expects. Ran it through docker build --target webapp-builder to make sure nx actually builds sq-server now, and it does.

Also agree on the CI point, build-ui never ran on this branch since it's a fork PR and needs to be approved first, that's on the repo side not something I can fix from here.

Went with keeping the webapp on 26.7 rather than falling back to 26.6 since 26.7 has the QP fix natively now and this addon fix covers the only thing that was actually broken.

Issue #1301, opened 2026-09-17
Forking and taking over maintainership


Magyarimiki:
hi,

I am wondering if it is time to fork this project and take over maintainership?
It seems @mc1arke is inactive for a while and now we have a couple of PRs in the queue to support new CE releases. Following the monthly release cycle is vital as using a non-matching plugin version is a no go due to the nature of the webapp.

Not being familiar with the codebase itself I might not be a good candidate to make this fork but I would be happy to test PRs when they are created.

I believe we would need at least 26.6 (last CE supporting Java 17) and the current 26.9 to be released.

mpet:
@Magyarimiki  I agree. Found out that we have made an internal fork within our company which is not good but I guess they did it for good reasons.

mc1arke:
> Following the monthly release cycle is vital 

Not without some level of support, it isn't. As with many open-source maintainers I have a life outside of open source: a job that pays the bills, family, hobbies, and other life commitments. Expecting a monthly release cadence - and the problem solving that comes with that for some releases - for free and to your timescales isn't a reasonable expectation. If you were contributing to the project (commits, MR reviews, triaging issues, donations etc) then I'd agree you have a reasonable expectation on response times, but stating an expectation from a project that you've not provided any support for - or provided support for any other projects if we were to consider reciprocal contribution for other software I obtain freely - isn't reasonable.

I'm still an active maintainer of this plugin: I'm triaging issues as they appear, but recent constraints on my time mean I've been unable to perform testing of release candidates, and am not prepared to release software I haven't verified.

Magyarimiki:
For _vital_ I meant that it is not a kind of plugin where you can use an older release for a non-matching server until the plugin of the same version is released. My organization is testing CE 26.9 and the missing plugin release is a showstopper so currently we go with the PR snapshot.

I completely respect your non-opensource life (I also have my own) and have no expectation towards you. Do not consider this to be an offense. I was just telling that currently the plugin is lagging a couple of versions behind Sonarqube while it seems there is an active community for creating and testing PRs. On the other hand,  as it can be seen from comments, users are waiting for updated releases.

Btw, I **did** contribute by testing PRs and reporting issues for them and planning to keep doing so.

brother:
Would be great to see some widening of the maintainership in the project. If some of the people around that has shown skills to make the updates could be invited as collaborators I think we can see something good happening.
[...]

PR #1298, opened 2026-09-08
Add SonarQube 26.9 support

Adrian-Eckardt:
## Summary

- update the SonarQube server API and Docker base image to `26.9.0.129388`
- update the webapp submodule to the `sqcb-26.9.0` tag
- restore addon module resolution under the SonarQube 26.9 Vite configuration
- replace the removed `HelpTooltip` component with the supported legacy `Tooltip`
- keep Docker builds compatible with the legacy builder while retaining automatic plugin upgrades in persistent extension volumes

This is built on top of #1296 because SonarQube 26.9 requires the 26.7/26.8 compatibility changes and persistent-volume plugin upgrade handling introduced there. Please merge #1296 first; after that, this PR's 26.9-specific commit can be reviewed on its own.

## Verification

- `./gradlew clean test`
- `./docker/test-community-branch-entrypoint.sh`
- production webapp build with `yarn nx run sq-server:build`
- native amd64 Docker image build for `26.9.0.129388-community`
- GitHub Actions build: https://github.com/Adrian-Eckardt/sonarqube-community-branch-plugin/actions/runs/34200462051
- disposable SonarQube 26.9 + PostgreSQL runtime tests covering main, named branch, and pull-request scans
- pull-request APIs verified for branch metadata, new-code measures, issues, and quality-gate status
- negative quality-gate test verified that `sonar.qualitygate.wait=true` returns scanner exit code 3
- deployed to an existing SonarQube 26.9 instance and verified a real GitLab merge-request analysis against `main` without creating a separate SonarQube project

PR #1296, opened 2026-08-31
Add SonarQube 26.8 support

Nacho-Larreta:
> [!IMPORTANT]
> Depends on #1280. This branch is stacked directly on the exact head commit of that PR and preserves its original commits and authorship. Once #1280 merges, this branch will be rebased onto `master` so this PR shows only the 26.8 follow-up.

## Summary

- add support for SonarQube Community Build 26.8.0.126808 and Java 21
- update the SonarQube webapp submodule to `sqcb-26.8.0`
- adapt moved overview helpers and Echoes `Spinner` usage for the 26.8 webapp
- make plugin installation resilient to Docker image upgrades with persistent extension volumes
- migrate generic and older versioned plugin JARs to `disabled-plugins` before installing the current immutable JAR
- exercise upgrade and restart behavior in CI for both image variants

This also provides the 26.8 compatibility requested in #1290.

## Verification

- `./gradlew clean build` on JDK 21
- production webapp build (`yarn nx run sq-server:build`)
- entrypoint syntax and upgrade/restart regression test
- development and release Docker image builds
- persistent-volume migration smoke test from a previous plugin version
- live SonarQube 26.8 branch and pull-request analysis with GitHub quality-gate decoration
- independent adversarial review: P1 = 0, P2 = 0

PR #1302, opened 2026-09-22
PoC: automate SonarQube branch integration testing

mpet:
## Purpose

This is a proof of concept for automating part of the release-candidate
verification discussed in #1301.

It deliberately reuses the project's existing Dockerfile and Docker Compose
setup rather than introducing another test framework.

## What it tests

- Builds the candidate plugin and webapp into the SonarQube image.
- Starts SonarQube and PostgreSQL.
- Waits for SonarQube to report `UP`.
- Verifies that `communityBranchPlugin` is installed.
- Creates a disposable project and analysis token.
- Runs a real SonarScanner analysis with `sonar.branch.name=poc-branch`.
- Verifies that `poc-branch` is available through
  `/api/project_branches/list`.
- Prints SonarQube logs on failure and cleans up the Docker resources.

## Result

The workflow has been tested successfully in my fork.

The SonarScanner analysis completed successfully and the integration test
confirmed that `poc-branch` was visible through the SonarQube API.

## Scope

This is intentionally a PoC rather than a proposal to make this a release
gate yet.

If the approach is useful, it could later be extended to pull-request
analysis and additional release-candidate checks.

Related to #1301.


joh-klein:
Thanks for starting this. Given @mc1arke's comment in #1301 that release-candidate testing is what's holding releases back, a harness that answers "does this candidate actually work?" seems like the most useful thing the community can add. I reviewed e39bc63 and ran it locally. Four suggestions follow, roughly in order of value. I'm happy to push these as commits to your branch if you'd rather.

**How I tested:** the released `mc1arke/sonarqube-with-community-branch-plugin:26.5.0.122743-community` image on Docker Desktop (arm64), running the script with `--no-build`. Building from source failed locally in `yarn install` (nx's postinstall on `node:22.16-alpine`, arm64), which I haven't looked into; CI on amd64 shouldn't be affected. I haven't run any of this on GitHub Actions.

### 1. The current check passes when the analysis fails

I rebuilt the plugin jar with `CommunityBranchLoaderDelegate.load()` throwing for `poc-branch`, a Compute Engine–side failure, and ran the script unchanged:

```
Branch 'poc-branch' is visible through the SonarQube API.
PoC integration test passed: the candidate plugin supports branch analysis.
```

`/api/ce/activity` for the same run showed `poc-branch FAILED deliberately broken for integration-test validation`. The branch row is created when the report is submitted (web side, `createBranchComponent`), so `project_branches/list` shows it whether or not the analysis succeeds. The scanner also prints `ANALYSIS SUCCESSFUL` as soon as the upload finishes, so its exit code doesn't tell you either.

The fix is to wait on the analysis's own Compute Engine task and then check that a quality gate was computed. One snag: the scanner image sets `SCANNER_WORKDIR_PATH=/tmp/.scannerwork`, so `report-task.txt` stays inside the container. Taking the task ID from the scanner's output avoids that, and also avoids the uid mismatch you'd hit writing into the mounted folder on a runner:

```bash
wait_for_ce_task() {  # $1 = CE task id
  local task_id="$1" status
  deadline=$((SECONDS + 120))
  while (( SECONDS < deadline )); do
    status="$(curl --silent --show-error --fail -u "$token:" "${SONARQUBE_URL}/api/ce/task?id=${task_id}" \
      | python3 -c 'import json,sys; print(json.load(sys.stdin)["task"]["status"])')"
    case "$status" in
      SUCCESS) return 0 ;;
      FAILED|CANCELED) echo "CE task ${task_id} ended in ${status}" >&2; return 1 ;;
    esac
    sleep "$POLL_INTERVAL_SECONDS"
  done
  echo "CE task ${task_id} did not finish within 120s" >&2
  return 1
}

assert_quality_gate() {  # $1 = "branch=<name>" or "pullRequest=<key>"
  local gate
  gate="$(curl --silent --show-error --fail -u "$token:" \
    "${SONARQUBE_URL}/api/qualitygates/project_status?projectKey=${TEST_PROJECT_KEY}&$1" \
    | python3 -c 'import json,sys; print(json.load(sys.stdin)["projectStatus"]["status"])')"
  [[ "$gate" == OK || "$gate" == ERROR ]] || { echo "No quality gate computed for $1 (got ${gate})" >&2; return 1; }
}
```

With these changes, the same broken jar fails the run: `CE task 49b2de8b-… ended in FAILED`, exit 1. With the unmodified plugin, all three analyses below pass.

### 2. Cover the main branch and PR analysis, not only a named branch

PR decoration is the plugin's other half, and it goes through different code (`sonar.pullrequest.*`). Wrapping the scanner call in a function makes each case one line:

```bash
SCANNER_IMAGE="${SCANNER_IMAGE:-sonarsource/sonar-scanner-cli:12.2.0.4256_8.1.0}"

analyse() {
  # The scanner image keeps .scannerwork (and report-task.txt) inside the container,
  # so take the CE task id from the scanner's own output instead.
  local out
  out="$(docker run --rm --network host -v "$PWD/build/integration-test-project:/usr/src" \
    "$SCANNER_IMAGE" -Dsonar.host.url="${SONARQUBE_URL}" -Dsonar.token="$token" "$@" | tee /dev/stderr)"
  wait_for_ce_task "$(sed -n 's|.*/api/ce/task?id=||p' <<<"$out")"
}

analyse                                               # main branch first, the realistic baseline
analyse -Dsonar.branch.name="${TEST_BRANCH}"
assert_quality_gate "branch=${TEST_BRANCH}"
analyse -Dsonar.pullrequest.key=1 -Dsonar.pullrequest.branch=feature/x -Dsonar.pullrequest.base=main
assert_quality_gate "pullRequest=1"
```

This replaces the existing `docker run` line. The `project_branches/list` loop after it can stay but no longer carries any weight.

### 3. Pin the scanner image

`sonarsource/sonar-scanner-cli` without a tag means a scanner release can break or change the test while the plugin stays the same, which is the opposite of what a release gate should do. The `SCANNER_IMAGE` default above pins it and still allows an override.

### 4. Path filters (minor)

Every open version bump (#1280, #1296, #1298) also changes `build.gradle`, so the current filter triggers on all of them. The gap is a PR that bumps only `.env` (the SonarQube base image) and the `sonarqube-webapp` submodule, as a patch release might. `docker/**` (new in #1296/#1298) and `release.Dockerfile` are also missing:

```yaml
      - '.env'
      - '.gitmodules'
      - 'sonarqube-webapp'      # the submodule gitlink, so this is not a glob
      - 'docker/**'
      - 'release.Dockerfile'
```

---

**Follow-up, probably a separate PR:** testing #1298 found that the Quality Gates and General Settings pages failed to load even though the build passed (https://github.com/mc1arke/sonarqube-community-branch-plugin/pull/1298#issuecomment-5682511605). That's a front-end break, so no API check will catch it. One Playwright smoke test that loads those two pages plus a branch overview, and fails on a console error, would have caught it. I kept it out of this list so this PR stays small enough to merge quickly.

Once this lands, running it against #1298's head and posting the commit IDs and job output would give the maintainer a concrete candidate to verify.

